PR #18, opened 2017-04-11
Unhardcoded app launcher icon filetype

palob:
Since you're going to install the icon to a location compliant with [freedesktop.org standards](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html) (I take it that way, otherwise you'd have to specify the full path) you don't need to specify the icon file extension in the `.desktop` launcher. It will be found anyway.

This facilitates the use of non-PNG (e.g. SVG) app icon themes.
Cf. this [example `.desktop` launcher](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#example).

### DIFF
--- a/data/anbox.desktop
+++ b/data/anbox.desktop
@@ -1,7 +1,7 @@
 [Desktop Entry]
 Name=Anbox
 Exec=anbox run --rootfs=android-rootfs
-Icon=anbox.png
+Icon=anbox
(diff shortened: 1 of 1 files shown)


morphis:
LGTM, Thanks for this!

PR #29, opened 2017-04-12
WIP: Compile on non-ubuntu

oleid:
_Note_: This branch is not to be merged at the moment. I merely want to share my progress of experimentation on a place others can find it easily.

This is what I need to be able to compile anbox on ArchLinux. 

dbus-cpp and friends are not available there, hence I prepared dbus-cpp and properties-cpp as submodules (branches submodule) and added them to the project. Also, a few changes to -Werror and friends are included to actually build the submodules.

Since sailfish-os doesn't provide dbus-cpp and friends, too - AFAIK - it's probably of use there, too.

EDIT:
Google-test is now included as submodule; building is for me on archlinux nothing more than:

```
git clone git@github.com:oleid/anbox.git
git checkout compile-on-non-ubuntu

git submodule update --init --recursive
mkdir build  && cd build && cmake ..
```

PR #44, opened 2017-04-12
Fix compilation on 17.04.

vosst:


src/anbox/bridge/platform_api_skeleton.cpp
```diff
@@ -22,6 +22,8 @@
 #include "anbox/wm/window_state.h"
 #include "anbox/logger.h"
 
+#undef Status
```
morphis: Please use proper #ifdef Status .. #endif

morphis: https://github.com/anbox/anbox/pull/29/commits/9c221215a1d7f80ada1da1e59797a3b249fea03c looks good to me

vosst: Good catch, fixed.

PR #55, opened 2017-04-12
Move snapcraft.yaml back to our toplevel directory

morphis:


snapcraft.yaml
```diff
@@ -115,6 +119,8 @@ parts:
       - libgl1-mesa-glx
       - libsdl2-2.0-0
       - libsdl2-gfx-1.0-0
+    install: |
+      make test
```
vosst: To satisfy my own curiosity: wouldn't we need a separate make install if we choose to override?

morphis: The funny thing is, we don't override, its running after the actual installation step.

PR #57, opened 2017-04-12
Hide container-manager command and print out a warning when not called as root

morphis:


src/anbox/cmds/container_manager.cpp
```diff
@@ -51,6 +51,13 @@ anbox::cmds::ContainerManager::ContainerManager()
 
   action([&](const cli::Command::Context&) {
     try {
+      if (getuid() != 0) {
+        ERROR("You're not running the container-manager as root. Generally you don't");
```
vosst: Please format as: https://gist.github.com/vosst/b22aa893bdb2cac778db19ba08b214b7

src/anbox/cli.h
```diff
@@ -243,6 +243,9 @@ class Command : public DoNotCopyOrMove {
   /// @brief description returns a longer string explaining the command.
   virtual Description description() const;
 
+  /// @brief hidden returns if the command is hidden from the user or not.
+  virtual bool hidden() const;
```
vosst: I would rather prefer virtual Visibility visibility() const; with enum class Visibility { visible, hidden };. With that, ctor invocations are easier to read, e.g.: Command("test", "test", "test", Visibility::hidden); We could also extend to cases like Visibility::visible_if_root.

morphis: Let me rework that in a follow up PR. Want to get this out today and I am lazy :-)

src/anbox/cmds/container_manager.cpp
morphis: I did this before but that breaks the formatting as we have then

ERROR [test.cpp:100] You're not running the container-manager as root. Generally you don't
want to run the container-manager manually unless you're a developer
as it is started by the init system of your operating system.

PR #64, opened 2017-04-12
README.md: Fix snapd install instructions

feikname:
On Xubuntu 16.10, following the README.md instructions gives this message:
 > error: unknown flag `clasic'

**Note:**  yes, the message has this weird inconsistency of having a backtick at the start and a single quote at the end.

However, if the argument order is changed, it works just fine again.

My `snap --version` output, in case it might be helpful:
```text
feik@feik-Q500A:~$ snap --version
snap    2.23.6
snapd   2.23.6
series  16
ubuntu  16.10
kernel  4.8.0-46-generic
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ installer will perform all necessary steps. You can install it on a system
 providing support for snaps by running
 
 ```
-$ snap install --classic anbox-installer
+$ snap install anbox-installer --classic
 ```
 
 Alternatively you can fetch the installer script via


Nbsaw:
I have the same problem :cat: 
My solution is upgrade my snap .. 
```shell
sudo apt-get update
sudo apt-get upgrade
```
And then I can installed it

feikname:
@Nbsaw Thanks for the suggestion, but upgrading the system does not fix this problem, even by [enabling the proposed packages](https://wiki.ubuntu.com/Testing/EnableProposed). The only working solution is to change the argument order, as done on this pull request.

I did an full upgrade right now to confirm:
```text
feik@feik-Q500A:~$ snap --version
snap    2.24+17.04
snapd   2.24+17.04
series  16
ubuntu  17.04
kernel  4.10.0-19-generic
```

feikname:
Related: #89

morphis:
The error message in the PR description is interesting here:

> error: unknown flag `clasic'

 `clasic' is need wrong. It needs to be --classic Can you check with the right option name again? There should be no issue with the order.

feikname:
@morphis You're right, it was a typo. Sorry! 😅

PR #127, opened 2017-04-17
apt clean after package install in clean-build.sh

PeterDaveHello:


scripts/clean-build.sh
```diff
@@ -36,6 +36,8 @@ apt-get install -qq -y \
   pkg-config \
   protobuf-compiler
 
+apt-get clean
```
morphis: Why do we need this?

PeterDaveHello: Isn't it good to make the docker image smaller?

morphis: It is but we only create an ephemeral image here which isn't persisted anywhere so an `apt-get clean` will not have any effect. If we would have a `Dockerfile` which prepares a docker container for us with all necessary things preinstalled then it would but in this case it doesn't.

PeterDaveHello: Well, it's still helpful if someone build it without docker but just bare-metal?

morphis: Ok, lets add it.

PR #149, opened 2017-04-19
add debian support

subjectdenied:
add support for debian


morphis:
This needs to be reworked once https://github.com/anbox/anbox/pull/152 lands.

PR #152, opened 2017-04-20
Migrate most things the anbox-installer did into a debian package

morphis:
This will allow us to have more control over what ends up on the disk of our users and also allows to introduce changes which will automatically taken into account on a package upgrade. The plan is to reduce the amount of things we ship which the anbox-common package to a minimum once snaps have enough support for all things we need.

Packages build from this PR are available in the anbox staging ppa: https://launchpad.net/~morphis/+archive/ubuntu/anbox-support-staging

Also see https://github.com/anbox/anbox-installer/pull/25


hub2git:
@morphis,
I thumbed-up this post. I'll give a thumbs up to any progress on this beautiful project. :)

debian/rules
```diff
@@ -0,0 +1,40 @@
...
+%:
+	dh $@ --parallel --fail-missing --with systemd
+
+override_dh_auto_build:
+	cat data/anbox.service.in | sed s:@SNAP_MOUNT_DIR@:/snap/bin:g > data/anbox.service
+	cat data/anbox.conf.in | sed s:@SNAP_MOUNT_DIR@:/snap/bin:g > data/anbox.conf
+
+override_dh_install:
+	VERSION=$(shell dpkg-parsechangelog -SVersion)
+	install -d $(CURDIR)/debian/tmp/usr/src
```
ogra1: unless you want to make this buildable on HURD, you should use DEB_DESTDIR (should point to debian/tmp/) instead of $(CURDIR)/debian/tmp/ (equally for the rest below)

debian/rules
```diff
@@ -0,0 +1,40 @@
...
+VERSION=$(shell dpkg-parsechangelog -SVersion)
+
+# include /usr/share/dpkg/default.mk
+
+%:
+	dh $@ --parallel --fail-missing --with systemd
+
+override_dh_auto_build:
+	cat data/anbox.service.in | sed s:@SNAP_MOUNT_DIR@:/snap/bin:g > data/anbox.service
+	cat data/anbox.conf.in | sed s:@SNAP_MOUNT_DIR@:/snap/bin:g > data/anbox.conf
```
ogra1: is there no need to call "dh_build" at all ?

morphis: Not really.

debian/rules
morphis: Thanks! Will integrate that.

PR #191, opened 2017-04-26
Build debs also during our travis job

morphis:


scripts/clean-build.sh
```diff
@@ -40,14 +40,24 @@ apt-get clean
...
-# In cases where anbox comes directly from a checked out Android
-# build environment we miss some symlinks which are present on
-# the host and don't have a valid git repository in that case.
-git clean -fdx . || true
-git reset --hard || true
+cleanup() {
+  # In cases where anbox comes directly from a checked out Android
+  # build environment we miss some symlinks which are present on
+  # the host and don't have a valid git repository in that case.
+  git clean -fdx . || true
```
vosst: Could we be more strict here and check if we have a .git for example? || true might shadow actual issues over time.

scripts/clean-build.sh
```diff
@@ -40,14 +40,24 @@ apt-get clean
...
 mkdir build || rm -rf build/*
 cd build
 cmake ..
 make -j10
 make test
+
+cleanup
+
+apt-get install -y build-essential curl devscripts gdebi-core dkms dh-systemd
+apt-get install -y $(gdebi --quiet --apt-line ./debian/control)
```
vosst: I *think* this requires an additional build dependency for gdebi.

scripts/clean-build.sh
morphis: Done.

scripts/clean-build.sh
morphis: The apt-get line above installs gdebi-core which ships gdebi.

PR #213, opened 2017-05-02
Close connection when adbd connection is closed

morphis:
This will ensure the message processor is correctly terminated and
a new connection is established when the adb server process on the
host opens a connection to us again.


stbergmann:
Looks like it makes 'adb shell' work reliably for me (where it often failed in the past).

PR #214, opened 2017-05-03
Start session manager when not already running

morphis:
If an application is started and the session manager is not yet running the launcher will fork and start the session manager while showing a splash screen so that the user knows something is happening.


src/anbox/ui/splash_screen.cpp
```diff
@@ -0,0 +1,85 @@
...
+    const auto message = utils::string_format("Failed to create window: %s", SDL_GetError());
+    BOOST_THROW_EXCEPTION(std::runtime_error(message));
+  }
+
+  auto surface = SDL_GetWindowSurface(window_);
+  SDL_FillRect(surface, nullptr, SDL_MapRGB(surface->format, 0xee, 0xee, 0xee));
+  SDL_UpdateWindowSurface(window_);
+
+  auto renderer = SDL_CreateRenderer(window_, -1, SDL_RENDERER_ACCELERATED);
+  auto img = IMG_LoadTexture(renderer, "/snap/anbox/current/snap/gui/icon.png");
```
morphis: The path and the texture creation / image loading still needs improvements to handle errors and find the icon.png at the right location.

src/anbox/cmds/launch.cpp
```diff
@@ -69,51 +86,95 @@ anbox::cmds::Launch::Launch()
...
+        });
+
+        const auto exe_path = utils::process_get_exe_path(::getpid());
+        if (!fs::exists(exe_path)) {
+          ERROR("Can't find correct anbox executable to run. Found %s but does not exist", exe_path);
+          return EXIT_FAILURE;
+        }
+
+        try {
+          const auto flags = core::posix::StandardStream::empty; // core::posix::StandardStream::stdout | core::posix::StandardStream::stderr;
```
morphis: @vosst I didn't looked into the implementation of process-cpp yet but `empty` seems to do what I would expect from `stdoud | stderr`: Allowing the child to print everything on the parent stdout/stderr.

src/anbox/cmds/launch.cpp
```diff
@@ -69,51 +86,95 @@ anbox::cmds::Launch::Launch()
     auto bus = std::make_shared<core::dbus::Bus>(core::dbus::WellKnownBus::session);
     bus->install_executor(core::dbus::asio::make_executor(bus, rt->service()));
 
+    const auto snap_path = utils::get_env_value("SNAP");
+    if (!snap_path.empty()) {
+      const auto resource_path = utils::string_format("%s/usr/share/anbox", snap_path);
```
vosst: I would rather prefer const auto resource_path = fs::path(snap_path) / "usr" / "share" / "anbox"; with fs begin shorthand for boost::filesystem.

src/anbox/cmds/launch.cpp
vosst: Yup, that's correct. With empty, neither stdout nor stderr are redirected.

src/anbox/cmds/launch.cpp
```diff
@@ -69,51 +86,95 @@ anbox::cmds::Launch::Launch()
...
+        }
+
+        try {
+          const auto flags = core::posix::StandardStream::empty; // core::posix::StandardStream::stdout | core::posix::StandardStream::stderr;
+          auto child = core::posix::fork([&]() {
+            auto grandchild = core::posix::exec(exe_path, args, env, flags);
+            grandchild.dont_kill_on_cleanup();
+            return core::posix::exit::Status::success;
+          }, flags);
+          child.wait_for(core::posix::wait::Flags::untraced);
```
vosst: You might want to leave a comment here, explaining why waiting for the child (and not the grandchild) is perfectly fine.

src/anbox/cmds/launch.cpp
```diff
@@ -69,51 +86,95 @@ anbox::cmds::Launch::Launch()
...
         trap->stop();
       });
     });
 
     boost::asio::deadline_timer timer(rt->service());
     timer.expires_from_now(max_wait_timeout);
     timer.async_wait([&](const boost::system::error_code&) {
-      WARNING("Stop waiting as we're already waiting for too long. Something is wrong");
-      WARNING("with your setup and the container may have failed to boot.");
+      WARNING("Stopped waiting as we're already waited for too long. Something");
```
vosst: we've already waited for too long

src/anbox/cmds/launch.cpp
```diff
@@ -69,51 +86,95 @@ anbox::cmds::Launch::Launch()
...
+      WARNING("is wrong with your setup or the container has failed to boot.");
```
vosst: Feels like we should provide instructions for grabbing logs and debugging potential issues here. Might safe us quite some issues.

src/anbox/config.cpp
```diff
@@ -41,6 +41,10 @@ void anbox::SystemConfiguration::set_data_path(const std::string &path) {
   data_path = path;
 }
 
+void anbox::SystemConfiguration::set_resource_path(const std::string &path) {
```
vosst: path should be fs::path

src/anbox/ui/splash_screen.cpp
```diff
@@ -0,0 +1,95 @@
...
+    SDL_DestroyWindow(window_);
+}
+
+void SplashScreen::process_events() {
+  event_thread_running_ = true;
+  while (event_thread_running_) {
+    SDL_Event event;
+    while (SDL_WaitEventTimeout(&event, 100)) {
+      switch (event.type) {
+      case SDL_QUIT:
```
vosst: Seems like you forgot to either set event_thread_running_ to true or communicate the quit request to the application here.

src/anbox/cmds/launch.cpp
morphis: Done.

src/anbox/cmds/launch.cpp
morphis: Done

src/anbox/cmds/launch.cpp
morphis: Added a comment to explain where bugs can be filed. I will propose another PR soon which will add a few more debugging capabilities which I will link here as well.

src/anbox/config.cpp
morphis: Done

src/anbox/ui/splash_screen.cpp
morphis: We need to simply ignore SDL_QUIT here so I dropped the full switch block.

src/anbox/cmds/launch.cpp
```diff
@@ -19,19 +19,36 @@
...
+const boost::posix_time::seconds max_wait_timeout{240};
+const int max_restart_attempts{3};
+const std::chrono::seconds restart_interval{5};
+}
+
+bool anbox::cmds::Launch::try_launch_activity(const std::shared_ptr<dbus::stub::ApplicationManager> &stub) {
+  try {
+    DEBUG("Sending launch intent %s to Android ..", intent_);
+    stub->launch(intent_, graphics::Rect::Invalid, stack_);
+  } catch (std::exception &err) {
```
vosst: Please catch const std::exception& err.

vosst: You also might want to have a catch (...) { ... } to avoid exceptions from bubbling up.

Issue #217, opened 2017-05-03
"make install" doesn't install required helper libraries


berolinux:
anbox relies on various helper libraries that are contained in (and built with) the anbox source tree.
However, those libraries don't get installed by "make install".

The anbox binary requires libanbox-core.so, which in turn requires librenderControl_dec.so, libOpenGLESDispatch.so, libanbox-protobuf.so, libxdg.so, libemugl_common.so, libGLESv1_dec.so, libGLESv2_dec.so, libOpenglCodecCommon.so (none of which get installed).

The fix should be fairly simple, but I'm wondering what the intended behavior is: Should those helper libraries get installed or should they be linked statically? (I'd probably prefer the latter)

morphis:
> The fix should be fairly simple, but I'm wondering what the intended behavior is: Should those helper libraries get installed or should they be linked statically? (I'd probably prefer the latter)

All the libs you listed are by default statically linked and no .so should ever be built (only .a). What makes you thing that those .so files (which are never build) should be installed?

berolinux:
If I don't patch anything, all the libs in the list are built as shared libraries without soversion, not static libraries (.so, not .a). I'm using cmake 3.8.1; maybe a default setting has changed between cmake versions at some point.

Given the absence of static libraries, the anbox binary gets linked dynamically, causing it to not start unless those libraries are installed. Will send a pull request to make sure the libs get built as static libs regardless of cmake version.

berolinux:
Submitted pull request #221 to address the issue. Works here, and I don't see how it could break things for anyone else

morphis:
@berolinux If travis we should be ok.

PR #221, opened 2017-05-04
Make sure helper libraries get built as static libs

berolinux:
Make sure helper libraries get built as static libs regardless
of the cmake version being used.

This fixes issue #217

Signed-off-by: Bernhard Rosenkränzer <bero@lindev.ch>


morphis:
LGTM, thanks!

PR #241, opened 2017-05-08
cmds: session_manager: Provide an option to use any container

lag-linaro:
Normally Anbox will use the default container provided by the
Anbox Container Manager, but some users may wish to run their
own container.  Here we're adding a --use-own-container flag
which tells the session-manager not to interact (configure/
start) the Container Manager.  This allows the user to utilise
any other bespoke container of their choosing.  For instance,
this new feature was testing using Docker container running
the android.img provided by Anbox.

Signed-off-by: Lee Jones <lee.jones@linaro.org>

### DIFF
--- a/src/anbox/cmds/session_manager.cpp
+++ b/src/anbox/cmds/session_manager.cpp
@@ -109,6 +109,9 @@ anbox::cmds::SessionManager::SessionManager(const BusFactory &bus_factory)
   flag(cli::make_flag(cli::Name{"standalone"},
                       cli::Description{"Prevents the Container Manager from starting the default container (Experimental)"},
                       standalone_));
+  flag(cli::make_flag(cli::Name{"experimental"},
+                      cli::Description{"Allows users to use experimental features"},
+                      experimental_));
 
   action([this](const cli::Command::Context &) {
     auto trap = core::posix::trap_signals_for_process(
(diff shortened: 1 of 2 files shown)


vosst:
Thanks for the change, adding the ability to use docker to run the android container is a great idea. That being said, I think Anbox should take over ownership of the container, at least in terms of managing its lifecycle. For that to work, extending Anbox to be able to talk to docker for starting/stopping containers should be straightforward and we could rework the current PR to add a flag --container, taking an identifier along the lines of "android@docker". What do you think?

lag-linaro:
There is definitely an argument to provide the support you mention, however I believe using the Docker API to control the life cycle of the container to be a separate piece of work.  In my use-case I need to start the container using the command line and a Docker file, thus the purpose of this patch is to start the session-manager in something akin to an independent or service mode, where it starts and waits for a connection from whatever Android container that attempts to connect to it.

src/anbox/cmds/session_manager.cpp
```diff
@@ -106,6 +106,12 @@ anbox::cmds::SessionManager::SessionManager(const BusFactory &bus_factory)
   flag(cli::make_flag(cli::Name{"window-size"},
                       cli::Description{"Size of the window in single window mode, e.g. --window-size=1024,768"},
                       window_size_));
+  flag(cli::make_flag(cli::Name{"standalone"},
+                      cli::Description{"Prevents starting the Container Manager and default container (Experimental)"},
```
morphis: Actually its more "Prevents starting the Android container via the container manager" as its not really starting the container-manager, it's already started by systemd.

Experimental -> experimental

src/anbox/cmds/session_manager.cpp
```diff
@@ -106,6 +106,12 @@ anbox::cmds::SessionManager::SessionManager(const BusFactory &bus_factory)
...
+                      standalone_));
+  flag(cli::make_flag(cli::Name{"experimental"},
+                      cli::Description{"Allows users to use experimental features (--standalone)"},
```
morphis: -> Allow to use experimental features

Lets drop --standalone here as it's already marked as experimental above.

morphis:
This mixes with https://github.com/anbox/anbox/pull/254 , can you rebase on master and drop fixes from #254 ?

PR #244, opened 2017-05-09
Read host os-release file when running inside the snap

morphis:


src/anbox/cmds/system_info.cpp
```diff
@@ -94,14 +95,18 @@ class SystemInformation {
  private:
   void collect_os_info() {
     os_info_.snap_based = (getenv("SNAP") != nullptr);
```
vosst: While you are at it, you could replace the naked getenv here.

src/anbox/cmds/system_info.cpp
```diff
@@ -94,14 +95,18 @@ class SystemInformation {
...
+    // If we're running from within a snap the best we can do is to
+    // access the hostfs and read the os-release file from there.
+    if (os_info_.snap_based && fs::exists(host_os_release_path))
+        path = host_os_release_path;
+
+    // Double check that there aren't any permission errors when trying
+    // to access the file (e.g. because of snap confinement)
+    if (fs::exists(path)) {
       anbox::utils::EnvironmentFile os_release(os_release_path);
       os_info_.name = os_release.value("NAME");
```
vosst: Please define constants for the magic keys here.

src/anbox/cmds/system_info.cpp
```diff
@@ -94,14 +95,18 @@ class SystemInformation {
...
-    if (fs::exists(os_release_path)) {
+    fs::path path = os_release_path;
+    // If we're running from within a snap the best we can do is to
+    // access the hostfs and read the os-release file from there.
+    if (os_info_.snap_based && fs::exists(host_os_release_path))
+        path = host_os_release_path;
+
+    // Double check that there aren't any permission errors when trying
+    // to access the file (e.g. because of snap confinement)
+    if (fs::exists(path)) {
```
vosst: How do you handle the case of being unable to read the respective file?

src/anbox/cmds/system_info.cpp
```diff
@@ -94,14 +95,18 @@ class SystemInformation {
...
       anbox::utils::EnvironmentFile os_release(os_release_path);
```
vosst: This should be `path` now.

src/anbox/cmds/system_info.cpp
morphis: In that case the default value "n/a" will be used which is assigned in the struct below.

src/anbox/cmds/system_info.cpp
morphis: Done.

src/anbox/cmds/system_info.cpp
morphis: Done.

src/anbox/cmds/system_info.cpp
morphis: Done.

PR #251, opened 2017-05-10
Use a static IP configuration for Android instead of a full blown DHCP setup

morphis:
Maintaining a dnsmasq instance to have a DHCP server running gives a few problems. For our case it is enough to hand a static IP configuration to Android and let it use that one only. Android requires a binary configuration in /data/misc/ethernet/ipconfig.txt in a very specific format the IpConfigBuilder class implements.


src/anbox/android/ip_config_builder.cpp
```diff
@@ -0,0 +1,108 @@
...
+  writer.write_string_with_size(assignment_key);
+  writer.write_string_with_size(assignment_to_string(assignment_));
+
+  writer.write_string_with_size(link_address_key);
+  writer.write_string_with_size(link_.address);
+  writer.write_unsigned_long(link_.prefix_length);
+
+  writer.write_string_with_size(gateway_key);
+  writer.write_unsigned_long(0);
+  writer.write_unsigned_long(1);
```
morphis: Those two need explanation and constants.

src/anbox/container/lxc_container.cpp
```diff
@@ -76,6 +78,53 @@ void LxcContainer::setup_id_maps() {
...
+  //
+  // See http://androidxref.com/7.1.1_r6/xref/frameworks/base/core/java/android/net/IpConfiguration.java
+  // for more details of the IP configuration format used here.
+
+  android::IpConfigBuilder ip_conf;
+  ip_conf.set_version(android::IpConfigBuilder::Version::Version2);
+  ip_conf.set_assignment(android::IpConfigBuilder::Assignment::Static);
+  ip_conf.set_link_address("192.168.250.2", 24);
+  ip_conf.set_gateway("192.168.250.1");
+  ip_conf.set_dns_servers({"8.8.8.8"});
```
morphis: Will add proper constants for these. Maybe also make the DNS servers configurable.

src/anbox/android/ip_config_builder.h
```diff
@@ -0,0 +1,63 @@
...
+
+#include "anbox/common/binary_writer.h"
+
+#include <string>
+#include <vector>
+#include <cstdint>
+
+namespace anbox {
+namespace android {
+struct IpConfigBuilder {
```
morphis: Needs to be class instead of struct and will also implement a few unit tests for this one.

src/anbox/common/binary_writer.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include "anbox/common/binary_writer.h"
+
+#include <cstring>
+#include <stdexcept>
+
+#include <boost/endian/buffers.hpp>
+
+namespace {
+bool is_little_endian() {
+  std::uint32_t v = 1;
```
vosst: This could very well be static.

src/anbox/android/ip_config_builder.h
```diff
@@ -0,0 +1,66 @@
...
+  };
+
+  enum class Assignment {
+    Static,
+    DHCP,
+  };
+
+  IpConfigBuilder() = default;
+
+  std::size_t write(common::BinaryWriter &writer);
```
vosst: write can  be const.

src/anbox/common/binary_writer.h
```diff
@@ -0,0 +1,57 @@
...
+    Big,
+    Little,
+  };
+
+  explicit BinaryWriter(std::vector<std::uint8_t>::iterator begin_,
+                        std::vector<std::uint8_t>::iterator end_);
+
+  void set_byte_order(Order order);
+
+  void write_unsigned_short(std::uint16_t value);
```
vosst: I have been guilty of this myself, but we should rather be explicit here and say: write_uint16/_uint32 etc..

src/anbox/common/binary_writer.cpp
morphis: Done

src/anbox/android/ip_config_builder.h
morphis: Done

src/anbox/common/binary_writer.h
morphis: Done

PR #252, opened 2017-05-11
Print warning if user starts container manager manually

morphis:
This will tell users what they are doing wrong when they are running the container manager manually like a lot of people do as our bug reports show.

### DIFF
--- a/src/anbox/cmds/container_manager.cpp
+++ b/src/anbox/cmds/container_manager.cpp
@@ -62,7 +62,7 @@ anbox::cmds::ContainerManager::ContainerManager()
         WARNING("");
       }
 
-      if (getuid() != 0) {
+      if (geteuid() != 0) {
         ERROR("You are not running the container-manager as root. Generally you don't");
         ERROR("want to run the container-manager manually unless you're a developer");
         ERROR("as it is started by the init system of your operating system.");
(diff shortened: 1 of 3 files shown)


src/anbox/cmds/container_manager.cpp
```diff
@@ -48,11 +48,22 @@ anbox::cmds::ContainerManager::ContainerManager()
...
     try {
+      if (!daemon_) {
+        WARNING("You are running the container manager manually which is most likely not");
+        WARNING("what you want. The container manager is normally started by systemd or");
+        WARNING("or another init system. If you still want to run the container-manager");
+        WARNING("you can get rid of this warning by starting with the --daemon option.");
+        WARNING("");
+      }
+
       if (getuid() != 0) {
```
vosst: This should be geteuid()

src/anbox/cmds/container_manager.cpp
```diff
@@ -48,11 +48,22 @@ anbox::cmds::ContainerManager::ContainerManager()
...
                       privileged_));
+  flag(cli::make_flag(cli::Name{"daemon"},
+                      cli::Description{"Mark service as being started as systemd daemon"},
+                      daemon_));
 
   action([&](const cli::Command::Context&) {
     try {
+      if (!daemon_) {
+        WARNING("You are running the container manager manually which is most likely not");
+        WARNING("what you want. The container manager is normally started by systemd or");
```
vosst: duplicate 'or'

PR #257, opened 2017-05-12
Improve stability of the AdbMessageProcessor

morphis:


src/anbox/qemu/adb_message_processor.cpp
```diff
@@ -113,18 +125,15 @@ void AdbMessageProcessor::wait_for_host_connection() {
...
+    auto handshake = utils::string_format("%04x%s", message.size(), message.c_str());
     messenger->send(handshake.data(), handshake.size());
-  } catch (std::exception &) {
+  } catch (...) {
     // Try again later when the host adb service is maybe available
+    host_notify_timer_.cancel();
     host_notify_timer_.expires_from_now(default_adb_wait_time);
-    host_notify_timer_.async_wait(
-        [&](const boost::system::error_code &) { wait_for_host_connection(); });
+    host_notify_timer_.async_wait([&](const boost::system::error_code &) { wait_for_host_connection(); });
```
vosst: I think we should limit to 'this', we should also check the error code for 'aborted' and bail out.

src/anbox/qemu/adb_message_processor.cpp
```diff
@@ -50,7 +54,11 @@ AdbMessageProcessor::AdbMessageProcessor(
 
 AdbMessageProcessor::~AdbMessageProcessor() {
   state_ = closed_by_host;
+
+  host_notify_timer_.cancel();
   host_connector_.reset();
+
+  // Unlock our lock to bring down any waiting instance
   active_instance.unlock();
```
vosst: Please use a std::unique_lock<std::mutex> instead of manually unlocking the mutex. You can construct the lock with std::unique_lock<std::mutex> lock_{active_instance, std::defer_lock_t{}};

Minor niggle: active_instance -> active_instance_

hub2git:
How do we update our AnBox so that this pull-request (and all other fixes) will be in our Anbox?

PR #276, opened 2017-05-18
Fix build with clang

berolinux:
Fixes building anbox with clang 4.0 -- removes unused declarations,
makes sure a template is visible at first instantiation, makes sure
exception declarations match between headers and implementations

Signed-off-by: Bernhard Rosenkränzer <bero@lindev.ch>


morphis:
LGTM! Thanks for doing this!

PR #302, opened 2017-05-30
Fix build with GCC 7.1

EddieRingle:
From the Porting to [GCC 7][1] page:

> Several C++ Standard Library headers have been changed to no longer include
> the `<functional>` header. As such, C++ programs that used components defined
> in `<functional>` without explicitly including that header will no longer
> compile.

> Previously components such as std::bind and std::function were
> implicitly defined after including unrelated headers such as `<memory>`,
> `<future>`, `<mutex>`, and `<regex>`.
> Correct code should `#include <functional>` to define them.

[1]: https://gcc.gnu.org/gcc-7/porting_to.html

Signed-off-by: Eddie Ringle <eddie@ringle.io>


morphis:
Nice! Thanks for doing this. LGTM.

PR #308, opened 2017-06-04
android alias

supaplextor:
make install needs sudo, and I found a hint in another issue, so I created an alias.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ $ make
 A simple
 
 ```
-$ make install
+$ sudo make install
 ```
 
 will install the necessary bits into your system.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -172,6 +172,13 @@ Running Anbox from a local build requires a few more things you need to know
 about. Please have a look at the ["Runtime Setup"](docs/runtime-setup.md)
 documentation.
 
+An alias will save typing later.
```
morphis: I would prefer not to add the alias as a recommendation to the read as this is really the choice of the user. Also abbreviating "Android" as starting the appmgr activity is wrong. Can you drop this and leave just the "sudo" change from above in the PR?

girst:
modifying a user's bashrc is no good idea. (not only will it anger many, but it might not even have the desired effect; e.g last line is `cat > /dev/null <<'EOF'`, user is using `zsh`/`fish`, etc)

much better would be providing a shell script named for example `appmgr` and _optionally_ put it in `$PATH`. 

(sorry for the additional noise - and thanks for this very interesting project!)

PR #309, opened 2017-06-05
dkms: fix modules build with more kernel versions

tmn505:
The Makefile for ashmem and binder have this line:
KERNEL_SRC ?= /lib/modules/$(shell uname -r)/build
so dkms will always build those modules against running kernel,
even for other kernel version headers. With this commit dkms will
always provide the necessary kernel version.

### DIFF
--- a/kernel/ashmem/dkms.conf
+++ b/kernel/ashmem/dkms.conf
@@ -1,7 +1,7 @@
 PACKAGE_NAME="anbox-ashmem"
 PACKAGE_VERSION="1"
 CLEAN="make clean"
-MAKE[0]="make all KVERSION=$kernelver"
+MAKE[0]="make all KERNEL_SRC=/lib/modules/$kernelver/build"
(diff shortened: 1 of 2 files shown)


ryanleesipes:
Will this make the modules build against 4.10 (versions later than 4.8)?

morphis:
LGTM, can you answer @ryanleesipes question?

tmn505:
@ryanleesipes AFAIK it should build fine against 4.10 but doesn't with 4.11. The commit fixes situation when You have installed multiply kernel versions and have distribution that doesn't run `dkms_autoinstaller` by default. Here https://ptpb.pw/66Zc is example log output when dkms wants to build modules for kernel 4.9.30-1-lts but the variable mentioned in my previous comment, placed in `Makefile`, forces dkms to always build against currently running kernel (4.11.2-1-ARCH, which is known to fail, see #223 ).

PR #332, opened 2017-06-24
Cleanup unexpected Makefile

zhsj:
Signed-off-by: Shengjing Zhu <zsj950618@gmail.com>


morphis:
Can you explain why this is needed?

zhsj:
@morphis the first commit is to remove the Makefile, which shouldn't be there since we use cmake.

But you can find the CI failed with the first commit. The reason is the debhelper think the build system is cmake now and it tries to build it. Which is not the case here, the deb package here should not build the anbox, and it doesn't install cmake build deps.

So override the auto config in debhelper to get ride of the building to pass the CI.

zhsj:
Previously the CI didn't complain since the debhelper think the build system is gnu make(there's a Makefile). And the Makefile does nothing.

morphis:
LGTM, thanks!

PR #350, opened 2017-07-11
Remove duplicated suffix on version string

zhsj:
In CMakeLists.txt, it already defines:

if (ANBOX_VERSION_SUFFIX)
    set(ANBOX_VERSION "${ANBOX_VERSION}-${ANBOX_VERSION_SUFFIX}")
endif()

So there's no need to add suffix again in code

Signed-off-by: Shengjing Zhu <i@zhsj.me>


morphis:
LGTM, thanks for fixing this.

PR #354, opened 2017-07-12
Add support for rootfs mounts via squashfuse

morphis:
In some environments (containers) we don't have support for loopback devices and therefor the best choice is to use squashfuse. The implementation hands the mount work off to the mount program so we don't have to care about any implementations details of the squashfuse program and leave this entirely up to the existing infrastructure.

The container manager will automatically fallback to use squashfuse when it detects there is no support for loopback devices.

### DIFF
--- a/src/anbox/utils.cpp
+++ b/src/anbox/utils.cpp
@@ -27,6 +27,8 @@
 
 #include <fcntl.h>
 #include <mntent.h>
+#include <sys/stat.h>
+#include <sys/types.h>
 
 #include "anbox/utils.h"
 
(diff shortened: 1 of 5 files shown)


src/anbox/utils.cpp
```diff
@@ -193,5 +194,17 @@ bool is_mounted(const std::string &path) {
 
 }
 
+std::string find_program_on_path(const std::string &name) {
```
bergotorino: Naming is a little bit confusing, something like: get_full_path (or just "which" :) imho would be more descriptive about what is actually being returned.

morphis: `get_full_path` is misleading as it suggests just that it expands the given thing to a full path but actually this function searches for the right path. So `find_` as prefix is correct. Using `bool` would contradict the `find_` prefix. I initially added this function to then use the returned path as input for `exec` but stopped doing that in the final implementation. Lets leave this as is as a nice helper with a possible future use case.

bergotorino: could be find_full_path :) however given your rationale I will not argue, find_program_on_path is fine.

Issue #368, opened 2017-07-17
always need to download android.img when build anbox snap?


jackwish:
I have a self-built android.img with which i'd like to build anbox snap. However, with the command below under anbox root dir, snapcraft always wants to download android.img.
```
$ mkdir android-images
$ cp /path/to/android.img android-images/android.img
$ snapcraft
```

I checked the configuration @ https://github.com/anbox/anbox/blob/master/snapcraft.yaml#L69

```
      wget http://build.anbox.io/android-images/$IMAGE_PATH/$IMAGE_NAME.img
      mv $IMAGE_NAME.img $SNAPCRAFT_PART_INSTALL/android.img
```

Looks like this behavior is intended?

jackwish:
btw, when downloading, what about let server handles the image file mapping (android.img -> path/to/android.img) rather than hard code (https://github.com/anbox/anbox/blob/master/snapcraft.yaml#L53) the image path in `snapcraft.yaml`?

supaplextor:
If you browse http://build.anbox.io/android-images/2017 you can find numerous images. Some for amd64 and one for armhf. I'll let devs comment further.

david2896482:
You can edit snapcraft.yaml
Let snapcraft using your own image

jackwish:
PR as https://github.com/anbox/anbox/pull/385

lemrm:
Could somebody explain in detail, how to build a snap with my own android.img

PR #385, opened 2017-07-28
snap build: prefer local android.img rather than always download

jackwish:
Previously we always load android.img from server when building snap.
Now, build with <ANBOX>/android-images/android.img if there is one.


jackwish:
fix issue https://github.com/anbox/anbox/issues/368

morphis:
LGTM, thanks!

Issue #331, opened 2017-06-22
[suggestion] Add a back button to the title bar


JLIT0:
It would be great if a back button could be added to the title bar for Android applications, as is the case in ARC for Chrome OS and in some other Android emulators.  Currently, it's possible to go back by pressing 'Esc', but many users will not know this keyboard shortcut.

![microsoft-office-android-apps-on-asus-chromebook-flip-100667638-orig](https://user-images.githubusercontent.com/24968860/27434506-3674733e-5759-11e7-86cc-1469984c62c0.png)

morphis:
Interesting idea. Will require us to provide client side decorations in X11 and wayland environments, but definitely doable.

PR #396, opened 2017-08-02
Add support for client side decorations

morphis:
This goes together with https://github.com/anbox/platform_frameworks_base/pull/3 and will enable the caption bar Android provides for all windows we show on a Linux desktop. We use capabilities from SDL to detect user interaction with the caption bar. The statically defined sizes may need some tweaking here and there.

Contributes to #331

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -123,7 +123,7 @@ set(SOURCES
     anbox/graphics/buffer_queue.cpp
     anbox/graphics/buffered_io_stream.cpp
     anbox/graphics/gl_renderer_server.cpp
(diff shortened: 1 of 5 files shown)


gjsman:
Why hasn't this been merged yet?

morphis:
> Why hasn't this been merged yet?

Only reason is less time on my side to coordinate the merge with a change on the Android side.

luisdavim:
Please merge this.

timofonic:
@morphis What about requesting more devs to the team? The 22 others seem inactive. What about asking them if they are still interested in the project and remove them if not? Think of it as "house-cleaning" ;)

PR #397, opened 2017-08-04
binder: update to latest version and fix build on older kernels

### LINKED ISSUE #223: binder and ashmem modules included with anbox don't build with kernel 4.11
There have been various API changes in kernel 4.11 (zap_page_range parameter change, lock_task_sighand, ...) that prevent the binder and ashmem modules included with anbox from compiling.

Also, binder in 4.11 has been changed in a few ways (moving bits and pieces to binder_context) that make the anbox patch not apply directly.

morphis:
This backports the binder driver from the latest upstream linux
kernel tree as of v4.13-rc3-118-g19ec50a43 and enables a working
build on older kernels (e.g. 4.4).

The newer driver allows specification of multiple binder device
nodes but for the moment we only enable the default "binder" one.

Fixes #223

### DIFF
--- a/kernel/binder/Makefile
+++ b/kernel/binder/Makefile
@@ -1,4 +1,4 @@
-ccflags-y += -I$(src) -Wno-int-conversion
+ccflags-y += -I$(src) -Wno-int-conversion -DCONFIG_ANDROID_BINDER_DEVICES="\"binder\""
 obj-m := binder_linux.o
 binder_linux-y := deps.o binder.o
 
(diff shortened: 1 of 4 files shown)


stgraber:
Looks good to me. The multi-device mechanism is very rudimentary and not very solid but hopefully it can be improved...

timofonic:
@morphis binder does not work on drm-tip, I need it because GPU issues with stable Linux. Please look at https://github.com/anbox/anbox/issues/453

PR #412, opened 2017-08-18
added uninstall target to cmake to allow make uninstall operation

rmescandon:
allows:
`sudo make uninstall`


morphis:
LGTM, thanks!

PR #419, opened 2017-08-23
Recover on adb server killed

alfonsosanchezbeato:
This PR makes sure that the adb connection can recover after an "adb kill-server" command.


morphis:
LGTM, thanks!

Issue #410, opened 2017-08-16
Changing Display Size together with App install leaves anbox disfunctional


Krawinkel:
anbox system-info
version: 3
snap-revision: 48
os:
  name: Ubuntu
  version: 16.04.3 LTS (Xenial Xerus)
  snap-based: true
kernel:
  version: Linux version 4.10.0-30-generic (buildd@lcy01-17) (gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4) ) #34~16.04.1-Ubuntu SMP Wed Aug 2 02:13:56 UTC 2017
  binder: true
  ashmem: true
graphics:
  egl:
    vendor: Mesa Project
    version: 1.4 (DRI2)
    extensions:
      - EGL_CHROMIUM_sync_control
      - EGL_EXT_buffer_age
      - EGL_EXT_create_context_robustness
      - EGL_EXT_image_dma_buf_import
      - EGL_KHR_config_attribs
      - EGL_KHR_create_context
      - EGL_KHR_fence_sync
      - EGL_KHR_get_all_proc_addresses
      - EGL_KHR_gl_renderbuffer_image
      - EGL_KHR_gl_texture_2D_image
      - EGL_KHR_gl_texture_3D_image
      - EGL_KHR_gl_texture_cubemap_image
      - EGL_KHR_image
      - EGL_KHR_image_base
      - EGL_KHR_image_pixmap
      - EGL_KHR_no_config_context
      - EGL_KHR_reusable_sync
      - EGL_KHR_surfaceless_context
      - EGL_KHR_wait_sync
      - EGL_MESA_configless_context
      - EGL_MESA_drm_image
      - EGL_MESA_image_dma_buf_export
      - EGL_NOK_texture_from_pixmap
      - EGL_WL_bind_wayland_display
  gles2:
    vendor: Intel Open Source Technology Center
    vendor: OpenGL ES-CM 1.1 Mesa 17.0.7
    extensions:
      - GL_EXT_blend_minmax
      - GL_EXT_multi_draw_arrays
      - GL_EXT_texture_filter_anisotropic
      - GL_EXT_texture_lod_bias
      - GL_OES_byte_coordinates
      - GL_OES_fixed_point
      - GL_OES_stencil_wrap
      - GL_OES_compressed_paletted_texture
      - GL_OES_query_matrix
      - GL_OES_read_format
      - GL_OES_single_precision
      - GL_EXT_texture_compression_dxt1
[...]

Issue #376, opened 2017-07-21
Anbox can not run


MonikaKoutna:
Make sure you are running the latest version of Anbox before reporting an issue.

Please also check that no similar bug is already reported. Have a look on the list of open bugs at https://github.com/anbox/anbox/issues

** Please paste the result of `anbox system-info` below:**
```
[version: 3
snap-revision: 44
os:
  name: Linux Mint
  version: 18.2 (Sonya)
  snap-based: true
kernel:
  version: Linux version 4.8.0-53-generic (buildd@lgw01-56) (gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4) ) #56~16.04.1-Ubuntu SMP Tue May 16 01:18:56 UTC 2017
  binder: true
  ashmem: true
graphics:
  egl:
    vendor: Mesa Project
    version: 1.4 (DRI2)
    extensions:
      - EGL_CHROMIUM_sync_control
      - EGL_EXT_create_context_robustness
      - EGL_EXT_image_dma_buf_import
      - EGL_KHR_create_context
      - EGL_KHR_fence_sync
      - EGL_KHR_get_all_proc_addresses
      - EGL_KHR_gl_renderbuffer_image
      - EGL_KHR_gl_texture_2D_image
      - EGL_KHR_gl_texture_cubemap_image
      - EGL_KHR_image
      - EGL_KHR_image_base
      - EGL_KHR_image_pixmap
      - EGL_KHR_reusable_sync
      - EGL_KHR_surfaceless_context
      - EGL_KHR_wait_sync
      - EGL_MESA_configless_context
      - EGL_MESA_drm_image
      - EGL_MESA_image_dma_buf_export
      - EGL_NOK_swap_region
      - EGL_NOK_texture_from_pixmap
      - EGL_NV_post_sub_buffer
      - EGL_WL_bind_wayland_display
  gles2:
    vendor: Intel Open Source Technology Center
    vendor: OpenGL ES-CM 1.1 Mesa 12.0.6
    extensions:
      - GL_EXT_blend_minmax
      - GL_EXT_multi_draw_arrays
      - GL_EXT_texture_filter_anisotropic
      - GL_EXT_texture_lod_bias
      - GL_OES_byte_coordinates
      - GL_OES_fixed_point
      - GL_OES_stencil_wrap
      - GL_OES_compressed_paletted_texture
      - GL_OES_query_matrix
      - GL_OES_read_format
      - GL_OES_single_precision
[...]

Issue #375, opened 2017-07-20
Anbox won't run 


MohamedxSalah:
Make sure you are running the latest version of Anbox before reporting an issue.

Please also check that no similar bug is already reported. Have a look on the list of open bugs at https://github.com/anbox/anbox/issues

** Please paste the result of `anbox system-info` below:**
```
version: 3
snap-revision: 44
os:
  name: Ubuntu
  version: 17.04 (Zesty Zapus)
  snap-based: true
kernel:
  version: Linux version 4.10.0-28-generic (buildd@lgw01-11) (gcc version 6.3.0 20170406 (Ubuntu 6.3.0-12ubuntu2) ) #32-Ubuntu SMP Fri Jun 30 05:32:18 UTC 2017
  binder: true
  ashmem: true
graphics:
  egl:
    vendor: Mesa Project
    version: 1.5 (DRI2)
    extensions:
      - EGL_CHROMIUM_sync_control
      - EGL_EXT_buffer_age
      - EGL_EXT_create_context_robustness
      - EGL_EXT_image_dma_buf_import
      - EGL_KHR_cl_event2
      - EGL_KHR_create_context
      - EGL_KHR_fence_sync
      - EGL_KHR_get_all_proc_addresses
      - EGL_KHR_gl_colorspace
      - EGL_KHR_gl_renderbuffer_image
      - EGL_KHR_gl_texture_2D_image
      - EGL_KHR_gl_texture_3D_image
      - EGL_KHR_gl_texture_cubemap_image
      - EGL_KHR_image
      - EGL_KHR_image_base
      - EGL_KHR_image_pixmap
      - EGL_KHR_reusable_sync
      - EGL_KHR_surfaceless_context
      - EGL_KHR_wait_sync
      - EGL_MESA_configless_context
      - EGL_MESA_drm_image
      - EGL_MESA_image_dma_buf_export
      - EGL_NOK_texture_from_pixmap
      - EGL_WL_bind_wayland_display
  gles2:
    vendor: X.Org
    vendor: OpenGL ES-CM 1.1 Mesa 12.0.6
    extensions:
      - GL_EXT_blend_minmax
      - GL_EXT_multi_draw_arrays
      - GL_EXT_texture_filter_anisotropic
      - GL_EXT_texture_lod_bias
      - GL_OES_byte_coordinates
      - GL_OES_fixed_point
      - GL_OES_stencil_wrap
      - GL_OES_compressed_paletted_texture
      - GL_OES_query_matrix
[...]

Issue #367, opened 2017-07-16
Anbox remains "starting" but never actually starts


definha:
Make sure you are running the latest version of Anbox before reporting an issue.

Please also check that no similar bug is already reported. Have a look on the list of open bugs at https://github.com/anbox/anbox/issues

** Please paste the result of `anbox system-info` below:**
defa@Defa:~$ anbox version
anbox 3
defa@Defa:~$ anbox system-info
version: 3
snap-revision: 44
os:
  name: Ubuntu
  version: 16.04.2 LTS (Xenial Xerus)
  snap-based: true
kernel:
  version: Linux version 4.10.0-041000-generic (kernel@gloin) (gcc version 6.2.0 20161005 (Ubuntu 6.2.0-5ubuntu12) ) #201702191831 SMP Sun Feb 19 23:33:19 UTC 2017
  binder: true
  ashmem: true
graphics:
  egl:
    vendor: Mesa Project
    version: 1.4 (DRI2)
    extensions:
      - EGL_CHROMIUM_sync_control
      - EGL_EXT_image_dma_buf_import
      - EGL_KHR_create_context
      - EGL_KHR_fence_sync
      - EGL_KHR_get_all_proc_addresses
      - EGL_KHR_gl_renderbuffer_image
      - EGL_KHR_gl_texture_2D_image
      - EGL_KHR_gl_texture_cubemap_image
      - EGL_KHR_image
      - EGL_KHR_image_base
      - EGL_KHR_image_pixmap
      - EGL_KHR_reusable_sync
      - EGL_KHR_surfaceless_context
      - EGL_KHR_wait_sync
      - EGL_MESA_configless_context
      - EGL_MESA_drm_image
      - EGL_MESA_image_dma_buf_export
      - EGL_NOK_swap_region
      - EGL_NOK_texture_from_pixmap
      - EGL_NV_post_sub_buffer
      - EGL_WL_bind_wayland_display
  gles2:
    vendor: Intel Open Source Technology Center
    vendor: OpenGL ES-CM 1.1 Mesa 12.0.6
    extensions:
      - GL_EXT_blend_minmax
      - GL_EXT_multi_draw_arrays
      - GL_EXT_texture_filter_anisotropic
      - GL_EXT_texture_lod_bias
      - GL_OES_byte_coordinates
      - GL_OES_fixed_point
      - GL_OES_stencil_wrap
      - GL_OES_compressed_paletted_texture
      - GL_OES_query_matrix
[...]

PR #422, opened 2017-08-24
Fix various problems with the session manager startup process

### LINKED ISSUE #416: Don't running
Make sure you are running the latest version of Anbox before reporting an issue.

Please also check that no similar bug is already reported. Have a look on the list of open bugs at https://github.com/anbox/anbox/issues

** Please paste the result of `anbox system-info` below:**

When i run anbox system-info, i have Segmentation error

**Please describe your problem:**

While running Anbox, i have a wite window for a secod, then it closing

**What were you expecting?:**

**Additional info:**

Ubuntu 17.04, Intel Pentium B980, Intel HD Graphics+Nvidia GT 620M, 4GB DDR3 RAM

E-mail for connect with me zaqqq13@gmail.com

morphis:
This improves the session manager startup process quite a lot.  Changes in detail are

 * Don't force host GL driver on nvidia platforms anymore; we don't need this as we're using the host GL driver by default now
 * Only attempt to create a SDL renderer when we don't have one already; a SDL window attempts to create a renderer on it's own at times so we check for that and only create one when necessary
 * Print out example launch intent when executed without any argument: people were confused what kind of arguments to provide to the launch command so we're now giving them an example
 * Apply proper daemonization handling for forked session manager: the most important change of this PR. We were lazy in forking of the session manager before and didn't ensured it's properly getting its own process group and becoming the leader of it. This fixes this and also redirects stdout/stderr/stdin to /dev/null so that subsequent users of these are not crashing when they are point to an invalid file descriptor
 * Correct locking when we draw a new frame: Operations inside the Renderer need to take a lock when operating on the EGL context to prevent EGL_BAD_ACCESS errors when eglMakeCurrent is called (context is used in another thread). This now locks correctly when a new frame is drawn.
 * RPC/DBus timeout improvements: We're now using a larger dbus call timeout and also give a timeout of five seconds for RPC calls to the Android container to keep the system operational even when something fails.

Fixes #416, #410, #404, #395, #392, #390, #387, #382, #376, #375, #367, #365, #344, #352, #343, #336, #335, #322, #317, #316, #285, #290, #310, #282, #275, #269 (and a lot more)

### DIFF
--- a/src/anbox/bridge/android_api_stub.cpp
+++ b/src/anbox/bridge/android_api_stub.cpp
@@ -29,6 +29,10 @@
 
 namespace fs = boost::filesystem;
 
+namespace {
+constexpr const std::chrono::milliseconds default_rpc_call_timeout{5000};
+} // namespace
+
 namespace anbox {
 namespace bridge {
 AndroidApiStub::AndroidApiStub() {}
@@ -103,7 +107,9 @@ void AndroidApiStub::launch(const android::Intent &intent,
       google::protobuf::NewCallback(this, &AndroidApiStub::application_launched,
                                     c.get()));
 
-  launch_wait_handle_.wait_for_all();
+  launch_wait_handle_.wait_for_pending(default_rpc_call_timeout);
+  if (!launch_wait_handle_.has_result())
+    throw std::runtime_error("RPC call timed out");
 
   if (c->response->has_error()) throw std::runtime_error(c->response->error());
 }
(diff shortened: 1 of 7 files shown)


the1jgroup:
Did this ever solve the perpetual Starting launch fail bug? Mine still happens.

512yottabytes:
Mine still happens too.

Fuseteam:
might be a different cause then was address, we would need to triage to figure that out

PR #423, opened 2017-08-24
cmds: directly start appmgr in single window mode

morphis:


morphis:
@rmescandon I agree but it should be sufficient for now. We can improve later if we have to.

PR #441, opened 2017-09-01
Read and set Lxc configuration from file 

jackwish:
Users  can define configurations like CPU cores for Lxc.

### DIFF
--- a/src/anbox/cmds/container_manager.cpp
+++ b/src/anbox/cmds/container_manager.cpp
@@ -52,6 +52,9 @@ anbox::cmds::ContainerManager::ContainerManager()
   flag(cli::make_flag(cli::Name{"daemon"},
                       cli::Description{"Mark service as being started as systemd daemon"},
                       daemon_));
+  flag(cli::make_flag(cli::Name{"lxc-conf"},
+                      cli::Description{"Path to Lxc configuration which manages container resource"},
+                      lxc_conf_path_));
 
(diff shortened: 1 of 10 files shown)


src/anbox/cmds/container_manager.cpp
```diff
@@ -63,6 +66,14 @@ anbox::cmds::ContainerManager::ContainerManager()
         WARNING("");
       }
 
+      if (!lxc_conf_path_.empty()) {
+        SystemConfiguration::instance().set_lxc_conf_path(lxc_conf_path_);
```
morphis: I don't like this very much. Anbox is meant to abstract the container engine from the user. Providing a path to an existing LXC configuration file is something only useful for a developer. If we want additional features like memory or CPU limits we should implement this in a generic way so the LXC backend can write the right bits into the configuration file.

What we can start with is adding static CPU/memory limits and then add a configuration file for Anbox which allows users to change the limits if they want. Something like `anbox config set limits.cpus 4` or `anbox config set limits.memory 4G`

What do you think?

jackwish: yeah, you are right. I had considered your proposal as well before coding. but... i am just too lazy to write all of them...

jackwish:
this design is not end-user-friendly. Developers or Linux expert users can refer to https://github.com/jackwish/anbox/tree/feature/lxc-config if needed.

PR #442, opened 2017-09-01
add file log for debug

Vichardw:
add log to files, so we can check Anbox log easily.


src/anbox/logger.cpp
```diff
@@ -178,25 +194,25 @@ void Logger::Fatal(const std::string& message,
 std::ostream& operator<<(std::ostream& strm, anbox::Logger::Severity severity) {
   switch (severity) {
     case anbox::Logger::Severity::kTrace:
-      return strm << "TT";
```
morphis: Please don't change these constants.

src/anbox/logger.cpp
```diff
@@ -178,25 +194,25 @@ void Logger::Fatal(const std::string& message,
...
+      return strm << "FATAL";
     default:
       return strm << static_cast<uint>(severity);
   }
 }
 
 std::ostream& operator<<(std::ostream& out, const Logger::Location& location) {
   return out << utils::string_format(
-             "%s:%d@%s",
+             "%s +%d::%s",
```
morphis: Why are you changing this?

src/anbox/logger.h
```diff
@@ -52,6 +52,7 @@ class Logger : public DoNotCopyOrMove {
   bool SetSeverityFromString(const std::string &severity);
   virtual void SetSeverity(const Severity& severity) = 0;
   virtual Severity GetSeverity() = 0;
+  virtual void SetLogFile(const std::string& filename){log_file_name = filename;};
```
morphis: Please follow the code style and add implementation to the actual implementation class.

src/anbox/cmds/session_manager.cpp
```diff
@@ -131,6 +131,11 @@ anbox::cmds::SessionManager::SessionManager()
                       use_system_dbus_));
 
   action([this](const cli::Command::Context &) {
+
+    Log().SetLogFile(std::string("/var/log/anbox/session.txt"));
+    Log().SetSeverityFromString("debug");
+    INFO("======================Session Manager Log======================");
```
morphis: No need for this. The filename should already tell you which log this is.

src/anbox/cmds/session_manager.cpp
```diff
@@ -131,6 +131,11 @@ anbox::cmds::SessionManager::SessionManager()
                       use_system_dbus_));
 
   action([this](const cli::Command::Context &) {
+
+    Log().SetLogFile(std::string("/var/log/anbox/session.txt"));
+    Log().SetSeverityFromString("debug");
```
morphis: Nope, this is wrong. Log level is influenced by specifying ANBOX_LOG_LEVEL= on the command line. We can in addition add a command line argument `--log-level=<level>`.

src/anbox/cmds/session_manager.cpp
```diff
@@ -131,6 +131,11 @@ anbox::cmds::SessionManager::SessionManager()
                       use_system_dbus_));
 
   action([this](const cli::Command::Context &) {
+
+    Log().SetLogFile(std::string("/var/log/anbox/session.txt"));
```
morphis: Using a static path here is wrong. We have a set of paths influenced by the cmake configuration already in config.h.in

Also this needs to take care of the correct path when running inside the snap. Please see src/anbox/system_configuration.cpp for details.

Thinking more about this, the path should come from a command line argument handed by the user. `--log-file=<path>` is what it should be.

src/anbox/cmds/launch.cpp
```diff
@@ -155,6 +161,7 @@ anbox::cmds::Launch::Launch()
         try {
           auto flags = core::posix::StandardStream::empty;
           auto child = core::posix::fork([&]() {
+#ifdef DIRECT_LOG_TO_NULL
```
morphis: Why does this need to be a compile time decision?

src/anbox/cmds/launch.cpp
```diff
@@ -97,6 +99,10 @@ anbox::cmds::Launch::Launch()
                       use_system_dbus_));
 
   action([this](const cli::Command::Context&) {
+
+    Log().SetLogFile("/var/log/anbox/launch.txt");
+    INFO("======================Launch Log======================");
```
morphis: Not need for this.

src/anbox/cmds/launch.cpp
```diff
@@ -39,6 +39,7 @@ const boost::posix_time::seconds max_wait_timeout{240};
 const int max_restart_attempts{3};
 const std::chrono::seconds restart_interval{5};
 
+#ifdef DIRECT_LOG_TO_NULL
```
morphis: See comment below.

src/anbox/cmds/container_manager.cpp
```diff
@@ -70,6 +71,15 @@ anbox::cmds::ContainerManager::ContainerManager()
         return EXIT_FAILURE;
       }
 
+      std::string log_dir = "/var/log/anbox";
+      if (!fs::exists(log_dir)) {
+        fs::create_directories(log_dir);
```
morphis: Creating this only from the container manager is pretty racy. Which part of the code ever tries to write into this directory should attempt to create the directory. Also things running inside the user session (like session-manager and launch commands) should not log into /var/log but into $HOME.

Making things inside /var/log accessible for every user/group on the system is definitive wrong.

src/anbox/cmds/container_manager.cpp
```diff
@@ -54,6 +54,7 @@ anbox::cmds::ContainerManager::ContainerManager()
                       daemon_));
 
   action([&](const cli::Command::Context&) {
+
```
morphis: Please don't add a extra line here.

Issue #486, opened 2017-10-13
Ashmem won't install on Manjaro 4.14rc3-1


zergling-man:
I trust this is the latest version. I installed the `-git` package from AUR: `anbox local-df774db`

I have done a cursory search for similar issues, and found nothing.

`anbox system-info`:
```
version: local-df774db
os:
  name: Manjaro Linux
  version: 
  snap-based: false
kernel:
  version: Linux version 4.14.0-1-MANJARO (builduser@manjaro) (gcc version 7.2.0 (GCC)) #1 SMP PREEMPT Mon Oct 2 11:44:46 UTC 2017
  binder: true
  ashmem: false
graphics:
  egl:
    vendor: Mesa Project
    version: 1.5 (DRI2)
    extensions:
      - EGL_CHROMIUM_sync_control
      - EGL_EXT_buffer_age
      - EGL_EXT_create_context_robustness
      - EGL_EXT_image_dma_buf_import
      - EGL_EXT_image_dma_buf_import_modifiers
      - EGL_KHR_cl_event2
      - EGL_KHR_config_attribs
      - EGL_KHR_create_context
      - EGL_KHR_create_context_no_error
      - EGL_KHR_fence_sync
      - EGL_KHR_get_all_proc_addresses
      - EGL_KHR_gl_colorspace
      - EGL_KHR_gl_renderbuffer_image
      - EGL_KHR_gl_texture_2D_image
      - EGL_KHR_gl_texture_3D_image
      - EGL_KHR_gl_texture_cubemap_image
      - EGL_KHR_image
      - EGL_KHR_image_base
      - EGL_KHR_image_pixmap
      - EGL_KHR_no_config_context
      - EGL_KHR_reusable_sync
      - EGL_KHR_surfaceless_context
      - EGL_KHR_wait_sync
      - EGL_MESA_configless_context
      - EGL_MESA_drm_image
      - EGL_MESA_image_dma_buf_export
      - EGL_NOK_texture_from_pixmap
      - EGL_WL_bind_wayland_display
  gles2:
    vendor: X.Org
    vendor: OpenGL ES-CM 1.1 Mesa 17.2.2
    extensions:
      - GL_EXT_blend_minmax
      - GL_EXT_multi_draw_arrays
      - GL_EXT_texture_filter_anisotropic
      - GL_EXT_texture_lod_bias
      - GL_OES_byte_coordinates
      - GL_OES_fixed_point
      - GL_OES_stencil_wrap
[...]

PR #504, opened 2017-10-29
replace deprecated vfs_read with kernel_read

sarangnemo:
vfs_read is no longer exported from kernel 4.14 (see #486). A related issue with ashmem and ZFS was resolved by replacing vfs_read with kernel_read:
https://github.com/zfsonlinux/spl/issues/656


hacker1024:
This has fixed the issue for me as well, please merge.

morphis:
Does this work with older kernels as well?

sarangnemo:
I can't test exhaustively, but it also works for me with kernel 4.9.  According to [this](https://github.com/zfsonlinux/spl/pull/659), kernel_read should replace vfs_read from kernel 3.9.

timofonic:
@zergling-man @sarangnemo @vasishath @hacker1024  @morphis   The Pull Request doesn't seem applied upstream yet. WHY? No updates since Nov 2017. Are there any developer alive? What about assign commit privileges to more interested developers?

oopen:
What do you expect to merge ???
Pre 3.9 kernel distrib ?
But you "only" support up 14.04 ubuntu witch is 3.13 kernel !!!
2 lts version but not recent 17.10 ubuntu ....... lol
Plese merge, we need it for all up to date distro aka ubuntu 17.10 & archlinux
you live in past

@timofonic "What about assign commit privileges to more interested developers?" Good idea.

But it is a generic GIT problem, we need other democratical method, that git doen't provide today, like vote to merge ...
What append when mainterner die/not active ? multiple fork ? this will a serious issue for foss

PR #512, opened 2017-11-10
build anbox when the default python version is 3.x

fpemud:


morphis:
Needs a rebase before it can be merged

PR #521, opened 2017-11-17
anbox: qemu: qemud_message_processor: Handle multiple 'cat'ed messages

lag-linaro:
It is possible for more than one message to be delivered concurrently.
When this happens, messages are concatenated together (end-to-end).
Current code errors out if the total received data is larger than the
known header_size (4) + the body size (reported in the header).
However, when >=2 messages arrive at the same time, the total received
data will be far greater than this.

The current code base actually handles multiple messages perfectly
already, so all we need to do is remove the early return and
everything "just works".

Signed-off-by: Lee Jones <lee.jones@linaro.org>


morphis:
LGTM

PR #545, opened 2017-12-09
Update README (change development state to alpha)

ShalokShalom:


morphis:
What is your reason for changing this?

ShalokShalom:
Because the homepage shows its in alpha

timofonic:
@ShalokShalom I agree.

@morphis What do you think about this?

Issue #527, opened 2017-11-21
 File "scripts/gen-emugl-entries.py", line 131 invalid syntax


Gooberpatrol66:
Make sure you are running the latest version of Anbox before reporting an issue.

Please also check that no similar bug is already reported. Have a look on the list of open bugs at https://github.com/anbox/anbox/issues

** Please paste the result of `anbox system-info` below:**
```
[please paste printout of `anbox system-info` here]
```

**Please describe your problem:**
build error

**What were you expecting?:**
build completion

**Additional info:**
Scanning dependencies of target GMock
[  0%] Creating directories for 'GMock'
[  1%] No download step for 'GMock'
[  2%] No patch step for 'GMock'
[  2%] No update step for 'GMock'
[  2%] Performing configure step for 'GMock'
-- The CXX compiler identification is GNU 5.4.0
-- The C compiler identification is GNU 5.4.0
-- Check for working CXX compiler: /usr/x86_64-pc-linux-gnu/gcc-bin/5.4.0/c++
-- Check for working CXX compiler: /usr/x86_64-pc-linux-gnu/gcc-bin/5.4.0/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Check for working C compiler: /usr/x86_64-pc-linux-gnu/gcc-bin/5.4.0/gcc
-- Check for working C compiler: /usr/x86_64-pc-linux-gnu/gcc-bin/5.4.0/gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Found PythonInterp: /usr/bin/python (found version "3.4.5") 
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - not found
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
-- Configuring done
-- Generating done
-- Build files have been written to: /home/nathan/Downloads/anbox/anbox/build/gmock
[  2%] Performing build step for 'GMock'
Scanning dependencies of target gmock_main
[  9%] Building CXX object CMakeFiles/gmock_main.dir/gtest/src/gtest-all.cc.o
[...]

PR #577, opened 2018-01-04
try to make sure python2 is used

step21:
Makes it less likely that python 3 is used for the python 2 script
Also closes https://github.com/anbox/anbox/issues/527 and https://github.com/anbox/anbox/pull/512


morphis:
LGTM

PR #579, opened 2018-01-05
Python3

step21:
Saw somewhere that there were some issues with this. If this is wanted, would be great if you could include it. It runs on both Python 2 and 3. Everything builds and tests are fine (in case it is covered).
It might have to be rebased or something as their is this one odd commit below but I couldn't find the right way to do it quickly. So maybe first you can say if you want to include it at all.


morphis:
LGTM

PR #580, opened 2018-01-06
Add basic touchscreen support

mariogrip:
This adds basic touchscreen support


morphis:
LGTM but can you check this doesn't break the desktop version?

morphis:
Tested that this works well on the desktop.

PR #600, opened 2018-02-01
Just waiting 3 seconds when stop container.

hanpfei:
If container manager dies before session manager dying,
the session manager cannot exit if it waits for all.

Signed-off-by: Han Pengfei <hanpfei@gmail.com>

### DIFF
--- a/src/anbox/container/management_api_stub.cpp
+++ b/src/anbox/container/management_api_stub.cpp
@@ -24,6 +24,9 @@
 
 namespace anbox {
 namespace container {
+
+const std::chrono::milliseconds ManagementApiStub::stop_waiting_timeout{3000};
+
 ManagementApiStub::ManagementApiStub(
     const std::shared_ptr<rpc::Channel> &channel)
     : channel_(channel) {}
@@ -75,7 +78,9 @@ void ManagementApiStub::stop_container() {
   channel_->call_method("stop_container", &message, c->response.get(),
       google::protobuf::NewCallback(this, &ManagementApiStub::container_stopped, c.get()));
 
-  c->wh.wait_for_all();
+  // If container manager dies before session manager, the session manager
+  // cannot exit if it waits for all, so just wait for 3 seconds.
+  c->wh.wait_for_pending(stop_waiting_timeout);
 
   if (c->response->has_error()) throw std::runtime_error(c->response->error());
 }
--- a/src/anbox/container/management_api_stub.h
+++ b/src/anbox/container/management_api_stub.h
@@ -56,6 +56,7 @@ class ManagementApiStub : public DoNotCopyOrMove {
 
   mutable std::mutex mutex_;
   std::shared_ptr<rpc::Channel> channel_;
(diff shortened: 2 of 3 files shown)


src/anbox/qemu/adb_message_processor.cpp
```diff
@@ -74,10 +74,11 @@ void AdbMessageProcessor::advance_state() {
 
       if (state_ == closed_by_host) {
         host_connector_.reset();
-        return;
+      } else {
+        wait_for_host_connection();
       }
 
-      wait_for_host_connection();
+      lock_.unlock();
```
morphis: What is this actually fixing? That lock_ is kept lock is intentional.

src/anbox/qemu/adb_message_processor.cpp
```diff
@@ -74,10 +74,11 @@ void AdbMessageProcessor::advance_state() {
 
       if (state_ == closed_by_host) {
         host_connector_.reset();
-        return;
+      } else {
```
morphis: Drop brackets

src/anbox/container/management_api_stub.h
```diff
@@ -56,6 +56,7 @@ class ManagementApiStub : public DoNotCopyOrMove {
 
   mutable std::mutex mutex_;
   std::shared_ptr<rpc::Channel> channel_;
+  static const std::chrono::milliseconds stop_waiting_millis;
```
morphis: This has to go into a anonymous namespace inside management_api_stub.cpp

Furthermore the variable type is already encoding that we have milliseconds here so no need to repeat that in the variable name. stop_waiting_timeout is what you want here as a name

src/anbox/container/management_api_stub.cpp
```diff
@@ -75,7 +78,7 @@ void ManagementApiStub::stop_container() {
   channel_->call_method("stop_container", &message, c->response.get(),
       google::protobuf::NewCallback(this, &ManagementApiStub::container_stopped, c.get()));
 
-  c->wh.wait_for_all();
+  c->wh.wait_for_pending(stop_waiting_millis);
```
morphis: I know why but this deserves a comment in the code

src/anbox/qemu/adb_message_processor.cpp
hanpfei: I found that sometimes the process will be blocked here, so it cannot exit normally.

src/anbox/qemu/adb_message_processor.cpp
```diff
@@ -72,12 +72,12 @@ void AdbMessageProcessor::advance_state() {
...
-      if (state_ == closed_by_host) {
+      if (state_ == closed_by_host)
         host_connector_.reset();
-        return;
-      }
+      else
+        wait_for_host_connection();
 
-      wait_for_host_connection();
+      lock_.unlock();
```
morphis: As described before, this is still wrong.

Issue #608, opened 2018-02-08
Problem building anbox-modules-binder against 4.15.2


neroita:
Hi , I've a laptop that need latest 4.15 series kernel to work well ( touchpad hangs with previous kernel ).
I have a ubuntu 17.10 stock with 4.15.2 kernel , when I try to install anbox it hangs on compiling anbox-modules-binder .

When I build the module I get:

make -C /lib/modules/4.15.2-041502-lowlatency/build V=0 M=$PWD
make[1]: ingresso nella directory "/usr/src/linux-headers-4.15.2-041502-lowlatency"
  CC [M]  /usr/src/anbox-modules-binder-9.orig/deps.o
  CC [M]  /usr/src/anbox-modules-binder-9.orig/binder.o
In file included from ./include/linux/module.h:18:0,
                 from /usr/src/anbox-modules-binder-9.orig/binder.c:28:
/usr/src/anbox-modules-binder-9.orig/binder.c:140:39: error: initialization from incompatible pointer type [-Werror=incompatible-pointer-types]
 module_param_call(stop_on_user_error, binder_set_stop_on_user_error,
                                       ^
./include/linux/moduleparam.h:233:24: note: in definition of macro ‘module_param_call’
   { .flags = 0, .set = _set, .get = _get };  \
                        ^~~~
/usr/src/anbox-modules-binder-9.orig/binder.c:140:39: note: (near initialization for ‘__param_ops_stop_on_user_error.set’)
 module_param_call(stop_on_user_error, binder_set_stop_on_user_error,
                                       ^
./include/linux/moduleparam.h:233:24: note: in definition of macro ‘module_param_call’
   { .flags = 0, .set = _set, .get = _get };  \
                        ^~~~
cc1: some warnings being treated as errors
scripts/Makefile.build:316: set di istruzioni per l'obiettivo "/usr/src/anbox-modules-binder-9.orig/binder.o" non riuscito
make[2]: *** [/usr/src/anbox-modules-binder-9.orig/binder.o] Errore 1
Makefile:1514: set di istruzioni per l'obiettivo "_module_/usr/src/anbox-modules-binder-9.orig" non riuscito
make[1]: *** [_module_/usr/src/anbox-modules-binder-9.orig] Errore 2
make[1]: uscita dalla directory "/usr/src/linux-headers-4.15.2-041502-lowlatency"
Makefile:8: set di istruzioni per l'obiettivo "all" non riuscito
make: *** [all] Errore 2

[...]

PR #611, opened 2018-02-10
fix argument in binder_set_stop_on_user_error

mcku:
binder_set_stop_on_user_error second argument breaks build.
#608


morphis:
Is this specific to a particular kernel version? If yes, then we need to ifdef this change

PR #612, opened 2018-02-12
kernel specific patch -- fix argument in binder_set_stop_on_user_error

mcku:
changes the argument for kernel 4.15 and later.


morphis:
LGTM, thanks!

PR #616, opened 2018-02-14
Update qemud_message_processor.cpp

danFromTelAviv:
I get the following error when trying to compile:
void anbox::qemu::QemudMessageProcessor::send_header(const size_t&)’:
/home/dan/anbox/src/anbox/qemu/qemud_message_processor.cpp:73:55: error: format ‘%lx’ expects argument of type ‘long unsigned int’, but argument 4 has type ‘size_t {aka unsigned int}’ [-Werror=format=]
   std::snprintf(header, header_size + 1, "%04lx", size);

removing the l seems to help

### DIFF
--- a/src/anbox/qemu/qemud_message_processor.cpp
+++ b/src/anbox/qemu/qemud_message_processor.cpp
@@ -70,7 +70,7 @@ void QemudMessageProcessor::process_commands() {
 
 void QemudMessageProcessor::send_header(const size_t &size) {
   char header[header_size + 1];
-  std::snprintf(header, header_size + 1, "%04lx", size);
+  std::snprintf(header, header_size + 1, "%04x", size);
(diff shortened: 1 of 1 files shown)


modscleo4:
This is a x86-only issue, you can solve it by changing the line
``` c
std::snprintf(header, header_size + 1, "%04lx", size);
```
to
```c
#ifdef __i386__
std::snprintf(header, header_size + 1, "%04x", size);
#else
std::snprintf(header, header_size + 1, "%04lx", size);
#endif
```

orent:
#include <cstdint>

  std::snprintf(header, header_size + 1, “%04” PRIxPTR);

morphis:
Superseeded by https://github.com/anbox/anbox/pull/667

Issue #628, opened 2018-03-03
Anbox doesn't start - libpulsecommon-8.0.so: cannot open shared object file


jeffebido:
System Info:

i5-3337U CPU
8 GB RAM
Radeon HD 7550M
OS: elementary OS 0.4.1 Loki (based on Ubuntu 16.04.3 LTS)

Anbox doesn't start after successful installation.

```
[please paste printout of `anbox system-info` here]
```
/snap/anbox/71/usr/bin/anbox: error while loading shared libraries: libpulsecommon-8.0.so: cannot open shared object file: No such file or directory

harrycode:
Same problem :(

NemanjaMilosevic:
@morphis Can you address this bug?

DaMatrix:
can reproduce on fresh mint 18.3 install

MechanoRealist:
Confirm same error on Zorin OS.

jacksondm33:
This is a duplicate of #626 
Can we close this issue?

Issue #627, opened 2018-03-02
Anbox does not open due to missing library


JORGETECH:
I tried to open anbox after installing it under Kubuntu 17.10 but there seems to be a problem with a missing library. The next error appears after launching anbox:

`/snap/anbox/71/usr/bin/anbox: error while loading shared libraries: libpulsecommon-8.0.so: cannot open shared object file: No such file or directory`

I can't even provide a log of "anbox system-info" because of the missing library.

jeffebido:
Same problem here

harrycode:
Same

DaMatrix:
can reproduce on fresh mint 18.3 install

jacksondm33:
This should be fixed now.

harrycode:
uh oh

harrycode:
oh well

PR #632, opened 2018-03-04
scripts: explicitly set linker search path for libpulsecommon-8.0

### LINKED ISSUE #626: Can't open Anbox after install on Ubuntu 17.10 (libpulsecommon-8.0.so missing)
Trying to open from terminal with `~$ anbox` or `~$ anbox system-info` results in the same output:
`/snap/anbox/71/usr/bin/anbox: error while loading shared libraries: libpulsecommon-8.0.so: cannot open shared object file: No such file or directory`

I'd try to reinstall `libpulse0`, but the problem persists.

morphis:
Fixes #626 #627 #628


harrycode:
Yes!

PR #646, opened 2018-03-19
Fix security buffer overflows

LocutusOfBorg:
Hello, I fixed some possible security issues, and wrong returns to if statement.

Please review carefully, I didn't test them at all (and neither I do know if the printf parameters such as "name" still makes sense, or should just be dropped)


debian/control
```diff
@@ -2,8 +2,8 @@ Source: anbox
 Section: misc
 Priority: optional
 Maintainer: Simon Fels <morphis@gravedo.de>
-Build-Depends: debhelper (>= 7), dkms, dh-systemd
-Standards-Version: 3.9.7
```
morphis: Why is this bumped?  We're still supporting older platforms which may not have the latest version of the standard or debhelper

scripts/clean-build.sh
```diff
@@ -55,9 +55,9 @@ cleanup
 
 mkdir build || rm -rf build/*
 cd build
-cmake ..
-make -j10
-make test
+cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo
```
morphis: We want to use the default here and not RelWithDebInfo. If we want RelWithDebInfo as the default it has to go into CMakeLists.txt

debian/control
morphis: Hm, I was thinking debhelper was at that version for xenial. However thanks for pointing this out. Leaving it at "9" is right as we still want to support xenial.

scripts/clean-build.sh
morphis: You're right that RelWithDebInfo helps with debugging etc. However as clean-build.sh is used as CI entry point we should build always with the defaults here. Lets set RelWithDebInfo as default in CMakeLists.txt then.

LocutusOfBorg:
hello, I think I did them :)

PR #667, opened 2018-04-01
Use the `z` as print format for `size_t` variable instead of `l` to m…

sideeffffect:
…ake it platform independent

`l` is a format for `long int`. It accidentally worked on 64 bit architectures, because `long int` has the same length as `size_t` there.
It didn't work on 32 bit architectures, where `size_t` is shorter than `long int`.
Let's use `z`, because that's the proper format for `size_t` -- it works on any platform.


sideeffffect:
This PR deprecates PR #616

PR #765, opened 2018-06-21
Extend system-info output with a list of CPU features

morphis:
This will help us to easily learn about the CPU on a system and which features it supports when a user reports a bug. We have seen problems with older CPUs from 2011 which do not seem to support all features Android needs.


zhsj:
Maybe you can simply include the output of `cat /proc/cpuinfo`.

morphis:
@zhsj That might get pretty verbose. I prefer a clean way to check for cpu features and will add some static checks into the container startup as well to have a clear error message for people with older CPUs.

PR #775, opened 2018-06-25
Supporting privileged containers

gmistick:


scripts/container-manager.sh
```diff
@@ -58,6 +58,11 @@ start() {
 		EXTRA_ARGS="$EXTRA_ARGS --use-rootfs-overlay"
 	fi
 
+	privileged_container="$(snapctl get privileged-container.enable)"
```
morphis: Let's call this `container.privileged`

PR #807, opened 2018-07-08
Improper use of "didn't" in install doc file

xefdan3ms:
You should say:
"If you haven't logged into the Ubuntu Store yet"

Instead of :
"If you didn't logged into the Ubuntu Store yet"


morphis:
LGTM, thanks!

PR #809, opened 2018-07-09
fix no switch default and old-style-cast errors (gcc 10)

vaurelios:


zhsj:
hmm, `-Wno-error=old-style-cast` is already set in `external/backward-cpp/CMakeLists.txt`, and `-Wno-error=switch-default` is in `external/cpu_features/CMakeLists.txt`. Not sure why the compiler still complains, could you paste the log?

morphis:
@zhsj #810 seems to be related to this one and has logs.

zhsj:
no, #801 complains `-Werror=switch-default` in backward-cpp.

zhsj:
btw, if I understand correctly, the current development version of gcc is 9, not sure where gcc 10 comes from...

vaurelios:
@morphis I've opened a issue bombela/backward-cpp#122

PR #811, opened 2018-07-11
fix overlay lowerdir order

zhsj:
kernel document says:

Multiple lower layers can now be given using the the colon (":") as a
separator character between the directory names.  For example:

  mount -t overlay overlay -olowerdir=/lower1:/lower2:/lower3 /merged

As the example shows, "upperdir=" and "workdir=" may be omitted.  In
that case the overlay will be read-only.

The specified lower directories will be stacked beginning from the
rightmost one and going left.  In the above example lower1 will be the
top, lower2 the middle and lower3 the bottom layer.

The user directory should be on top, so that user can override the file
in original rootfs image.


morphis:
LGTM, thanks!

PR #865, opened 2018-08-16
Do application reset on first store_or_update

mariogrip:
This keep the desktop icons on anbox bootup until new icons gets added. 

This is mostly for ubuntu touch launcher


morphis:
LGTM

PR #888, opened 2018-09-02
Fix clang build errors

raetiacorvus:
This PR fixes two clang build errors. There are still three left but I was unsure how they should be handled since they are warnings for unused variables and i am not familiar with anbox codebase so i can not tell if they are false positives or not.

```
src/anbox/cmds/check_features.cpp:41:11: error: lambda capture 'this' is not used [-Wunused-lambda-capture]
          ^
…
src/anbox/dbus/bus.h:51:8: error: private field 'type_' is not used [-Wunused-private-field]
  Type type_;
       ^
…
/home/raetiacorvus/Documents/aur/anbox-git/src/anbox/src/anbox/graphics/emugl/RenderApi.cpp:33:23: error: unused variable 'default_glesv1_lib' [-Werror,-Wunused-const-variable]
constexpr const char *default_glesv1_lib{"libGLESv1_CM.so.1"};
                      ^
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,12 @@ set(C_AND_CXX_WARNINGS "-pedantic -Wall -Wextra")
 set(EXTRA_C_WARNINGS "-Wcast-align -Wcast-qual -Wformat -Wredundant-decls -Wswitch-default")
 set(EXTRA_CXX_WARNINGS "-Wnon-virtual-dtor -Wold-style-cast")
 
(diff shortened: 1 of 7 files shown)


raetiacorvus:
Ok looks like non gcc pragmas aren't allowed so what is this projects preferred way to handle this? Should I just downgrade the error to a warning by passing -Wno-error=c99-extensions globally for clang builds?

morphis:
@raetiacorvus It should be fine to drop the three mentioned unused variables.

PR #897, opened 2018-09-07
Update check_features.cpp - Added override for Pentium T4500

nvschilleman:


src/anbox/cmds/check_features.cpp
```diff
@@ -36,6 +36,9 @@ std::vector<std::string> cpu_whitelist = {
   "M 460",
   // Intel Celeron N2840
   "N2840",
+  // Intel Pentium T4500
+  "T4500",
+
```
morphis: Please drop the unneeded empty line.

PR #902, opened 2018-09-08
Improve performance

mariogrip:
This removes unnecessary std::copy 

For the amount these functions get called, std copy is way to expensive
to do. This removes the need for buffer to be copied twice, first to a
std::vector then SmallFixedVector. This also directly copies buffer into
a smallVector without the init overhead smallVector's ctor does. This
result is a massive performance boost (at least on arm devices).

Also use memcpy interestingly seems to be faster on arm
devices, on amd64 they seems to be about the same speed.

### DIFF
--- a/src/anbox/graphics/opengles_socket_connection.cpp
+++ b/src/anbox/graphics/opengles_socket_connection.cpp
@@ -1,0 +1,72 @@
+/*
+ * Copyright © 2018 The UBports project
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Marius Gripsgard <marius@ubports.com>
+ */
+
+#include "anbox/logger.h"
+
+#include "anbox/network/message_receiver.h"
+#include "anbox/network/message_sender.h"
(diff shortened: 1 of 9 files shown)


src/anbox/graphics/opengles_socket_connection.cpp
```diff
@@ -0,0 +1,72 @@
...
+  if (error) {
+    connections_->remove(id());
+    return;
+  }
+
+  Buffer data{};
+  data.resize_noinit(bytes_read);
+
+  // Use memcpy here, its faster then std::copy
+  memcpy(data.data(), buffer_.data(), bytes_read);
```
morphis: I think we should use memcpy in SocketConnection instead of adding an entirely new class inheriting from it. This is something every user of SocketConnection can benefit from.

src/anbox/network/message_processor.h
```diff
@@ -20,13 +20,15 @@
...
+#include "anbox/common/small_vector.h"
 
 namespace anbox {
 namespace network {
 class MessageProcessor {
  public:
   virtual ~MessageProcessor() {}
-  virtual bool process_data(const std::vector<std::uint8_t> &data) = 0;
+  virtual bool process_data(const std::vector<std::uint8_t> &) { return false; }
+  virtual bool process_data(anbox::common::SmallFixedVector<char, 512> &&) { return false; }
```
morphis: I would like to use a 

`use MessageBuffer = SmallFixedVector<chat, 512>;`

here

src/anbox/graphics/opengles_socket_connection.cpp
mariogrip: Problem was that SocketConnection creates a vector<> instead of SmallVector, so i created a new class that copies directly to a smallVector instead of copying to a vector then smallVector and i didnt want to change all the socketConnections to smallvector since std::vector has better dynamic memory management, so its not really beneficial to use that on other connections.

src/anbox/network/message_processor.h
mariogrip: Yeah, will add that

morphis:
@mariogrip Any updates on this?

PR #905, opened 2018-09-09
Use anbox-binder

mariogrip:
If /dev/anbox-binder is available, use that binder to not cause problem with existing binder.

This is mostly UBports specific.

### DIFF
--- a/src/anbox/cmds/session_manager.cpp
+++ b/src/anbox/cmds/session_manager.cpp
@@ -126,7 +126,13 @@ anbox::cmds::SessionManager::SessionManager()
       return EXIT_FAILURE;
     }
 
-    if (!fs::exists("/dev/binder") || !fs::exists("/dev/ashmem")) {
+    auto binder = "/dev/binder";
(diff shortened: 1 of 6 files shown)


src/anbox/cmds/session_manager.cpp
```diff
@@ -126,7 +126,13 @@ anbox::cmds::SessionManager::SessionManager()
       return EXIT_FAILURE;
     }
 
-    if (!fs::exists("/dev/binder") || !fs::exists("/dev/ashmem")) {
+    auto binder = "/dev/binder";
+    // If available, use own binder to not cause problem with existing binder
+    if (fs::exists("/dev/anbox-binder")) {
```
morphis: I think we should make the name of the binder node more a configuration option so you can pick for Ubports `-DBINDER_PATH=/dev/anbox-binder` otherwise we default to `/dev/binder`. That avoids having a list of binder nodes in the code which would just get larger if other projects pick a different name.

morphis:
@mariogrip Any updates on this one?

PR #907, opened 2018-09-09
Add option to disable sdl window hit test with envar value

mariogrip:


src/anbox/platform/sdl/window.cpp
```diff
@@ -72,8 +72,9 @@ Window::Window(const std::shared_ptr<Renderer> &renderer,
     BOOST_THROW_EXCEPTION(std::runtime_error(message));
   }
 
-  if (SDL_SetWindowHitTest(window_, &Window::on_window_hit, this) < 0)
-    BOOST_THROW_EXCEPTION(std::runtime_error("Failed to register for window hit test"));
+  if (utils::get_env_value("ANBOX_NO_SDL_WINDOW_HIT_TEST", "false") == "false")
```
morphis: I think for mobile platforms like Ubports we want to have a runtime switch for this which allows you to do something like `SetMode("mobile")` (we need to pick better names though) which then disables or optimizes a few things. For now this is fine.

PR #942, opened 2018-10-14
cmake: link gtest_main to gmock lib

zhsj:
Starting with googletest 1.8.1, I meet the following error when building
the tests.

```
/usr/bin/ld: ../../../gmock/libgmock_main.a(gmock_main.cc.o): in function `main':
/usr/src/googletest/googlemock/src/gmock_main.cc:52: undefined reference to `testing::InitGoogleMock(int*, char**)'
```

With this patch, I can build it successfully.

Build log: https://buildd.debian.org/status/logs.php?pkg=anbox&arch=amd64


morphis:
LGTM, thanks!

PR #943, opened 2018-10-14
Drop unneeded glib dependency

CameronNemo:
let me trigger your CI real quick and see what goes down

edit: well the docker build passed (neat!). y'all should try building the snap package too.


morphis:
LGTM

Issue #944, opened 2018-10-14
During snap install of anbox, received missing support for SSE 4.2 error


aherger:
1. Please check that no similar bug is already reported. Have a look on the list of open bugs at https://github.com/anbox/anbox/issues
2. Make sure you are running the latest version of Anbox before reporting an issue. Update snap to latest: `snap refresh --devmode --edge anbox`
3. Make sure you have debug logs enabled:
`sudo snap set anbox debug.enable=true`
4. Reproduce the error while debug logs enabled.
5. Run the anbox logs collection utility and attach the tar file.
`sudo /snap/bin/anbox.collect-bug-info `

6. ** Please paste the result of `anbox system-info` below:**
```
[please paste printout of `anbox system-info` here]
```

**Please describe your problem:**
Received following error while installing with snap:

2018-10-14T12:48:06-05:00 INFO Waiting for restart...
error: cannot perform the following tasks:
- Run install hook of "anbox" snap if present (run hook "install": 
-----
The CPU of your computer (Intel(R) Core(TM)2 Duo CPU     T9600  @ 2.80GHz) does not support all
features Anbox requires.
It is missing support for the following features: SSE 4.2
You can for example find more information about SSE
here https://en.wikipedia.org/wiki/Streaming_SIMD_Extensions

**What were you expecting?:**
I noticed a whitelist for certain CPU's in the check_features.cpp file.  Could you add this CPU to that whitelist, assuming that would take care of this issue?

-- Thanks

**Additional info:**

PR #967, opened 2018-11-03
Corrections/additions to "CPU whitelist"

Zjaw:
I believe a comma was forgotten in the array from my last pull request (line 46). I don't know if that makes a difference or not in C++, but it still wasn't installing for me. I tried to correct that issue, reorganized the list, and added another requested CPU [(#944)](https://github.com/anbox/anbox/issues/944).

### DIFF
--- a/src/anbox/cmds/check_features.cpp
+++ b/src/anbox/cmds/check_features.cpp
@@ -30,22 +30,22 @@ std::vector<std::string> cpu_whitelist = {
   // The following CPUs do not support AVX and without it cpu_features can't detect
   // if SSE & friends are supported. See https://github.com/google/cpu_features/issues/4
 
+  // Intel Core i7 Q720
+  "Q 720",
   // Intel Core i7 M620
   "M 620",
   // Intel Core i5 M460
   "M 460",
   // Intel Celeron N2840
   "N2840",
-  // Intel Core i7 Q720
-  "Q 720",
-  // Intel Pentium T4500
-  "T4500", 
-  // Intel Core i7 Q720
-  "Q 720",
   // Intel Xeon E5520
   "E5520",
+  // Intel Pentium T4500
+  "T4500",
   // Intel Core2 Duo T6500
-  "T6500"
+  "T6500",
+  // Intel Core2 Duo T9600
+  "T9600"
 };
 } // namespace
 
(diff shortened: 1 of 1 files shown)


src/anbox/cmds/check_features.cpp
```diff
@@ -30,22 +30,22 @@ std::vector<std::string> cpu_whitelist = {
...
   // Intel Xeon E5520
-  "E5520"
+  "E5520",
+  // Intel Pentium T4500
+  "T4500",
   // Intel Core2 Duo T6500
-  "T6500"
+  "T6500",
+  // Intel Core2 Duo T9600
+  "T9600"
```
morphis: According to http://www.cpu-world.com/CPUs/Core_2/Intel-Core%202%20Duo%20Mobile%20T9600%20AW80576GH0726M%20(BX80576T9600).html the CPU doesn't support SSE 4.2 so we can't add it here.

src/anbox/cmds/check_features.cpp
```diff
@@ -30,22 +30,22 @@ std::vector<std::string> cpu_whitelist = {
...
-  "T4500", 
-  // Intel Core i7 Q720
-  "Q 720",
   // Intel Xeon E5520
-  "E5520"
+  "E5520",
+  // Intel Pentium T4500
+  "T4500",
   // Intel Core2 Duo T6500
-  "T6500"
```
morphis: Actually we have to drop this one too as it doesn't support SSE 4.2. See http://www.cpu-world.com/CPUs/Core_2/Intel-Core%202%20Duo%20Mobile%20T6500%20AW80577GG0452MA%20-%20AW80577GG0452ML.html

Zjaw: So SSE 4.1 is no longer viable to run anbox?

morphis: No, as stated in the code you need SSE 4.2 too. Did you run Anbox before with just SSE 4.1?

Zjaw: I was never able to get it to install. I thought at some point I had seen a thread or documentation stating that SSSE 3 and SSE 4.1 were supported, but if that is not correct than I do apologize. I was just hoping that whitelisting my CPU was all that was needed to get me up and running.

morphis: Nope, whitelisting wont help as long as the CPU doesn't support the necessary instructions.

PR #985, opened 2018-11-23
dbus: fix stack smashing

raetiacorvus:
The expected C type of a sd-bus boolean is int and not boolean.
This fixes a possible stack smashing caused by the varying types.

See https://github.com/libratbag/libratbag/issues/433 which had the same
problem.

### DIFF
--- a/src/anbox/dbus/stub/application_manager.cpp
+++ b/src/anbox/dbus/stub/application_manager.cpp
@@ -40,7 +40,7 @@ ApplicationManager::ApplicationManager(const BusPtr& bus)
 ApplicationManager::~ApplicationManager() {}
 
 void ApplicationManager::update_properties() {
-  bool ready = false;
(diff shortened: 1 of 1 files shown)


src/anbox/dbus/stub/application_manager.cpp
```diff
@@ -40,7 +40,7 @@ ApplicationManager::ApplicationManager(const BusPtr& bus)
 ApplicationManager::~ApplicationManager() {}
 
 void ApplicationManager::update_properties() {
-  bool ready = false;
+  int ready = false;
```
morphis: Changing to `int` is fine but we should drop the use of `false` then and use `0`

raetiacorvus: Makes sense. Thanks for the review. Fixed it.

PR #1013, opened 2018-12-31
Sync mouse wheel events

awaken1ng:
Mouse wheel event doesn't fire off on Android side until, for example, you move the mouse, making scrolling very clunky.

This adds EV_SYN to mouse wheel handler to address that.

And since now it makes every mouse event handler add EV_SYN, I removed them from handlers and append it before sending the events.


morphis:
Thanks for fixing this!

PR #1030, opened 2019-01-19
Touch on linux, some fixes

salvprest:
before of SDL 2.0.7 on X11 tfinger coordinates are not normalized, Ubuntu 16.04 use SDL 2.0.4, starting from SDL 2.0.7, SDL normalize the tfinger coordinates in the private fuction xinput2_normalize_touch_coordinates.
The second commit point out that windowID is not a field of SDL_TouchFingerEvent event, and window is always NULL for touch events, so set windowID when the window is created or when the window take input focus

### DIFF
--- a/src/anbox/platform/sdl/platform.cpp
+++ b/src/anbox/platform/sdl/platform.cpp
@@ -268,7 +268,7 @@ void Platform::process_input_event(const SDL_Event &event) {
 
       window = SDL_GetWindowFromID(focused_sdl_window_id);
       // before SDL 2.0.7 on X11 tfinger coordinates are not normalized
-      if (!SDL_VERSION_ATLEAST(2,0,7) && (event.tfinger.x>1 || event.tfinger.y>1)) {
+      if (!SDL_VERSION_ATLEAST(2,0,7) && (event.tfinger.x > 1 || event.tfinger.y > 1)) {
         rel_x = event.tfinger.x;
         rel_y = event.tfinger.y;
       } else {
@@ -322,7 +322,7 @@ void Platform::process_input_event(const SDL_Event &event) {
 
       window = SDL_GetWindowFromID(focused_sdl_window_id);
       // before SDL 2.0.7 on X11 tfinger coordinates are not normalized
-      if (!SDL_VERSION_ATLEAST(2,0,7) && (event.tfinger.x>1 || event.tfinger.y>1)) {
+      if (!SDL_VERSION_ATLEAST(2,0,7) && (event.tfinger.x > 1 || event.tfinger.y > 1)) {
         rel_x = event.tfinger.x;
         rel_y = event.tfinger.y;
       } else {
--- a/src/anbox/platform/sdl/platform.cpp
+++ b/src/anbox/platform/sdl/platform.cpp
@@ -316,8 +316,8 @@ void Platform::process_input_event(const SDL_Event &event) {
 }
 
 bool Platform::calculate_touch_coordinates(const SDL_Event &event,
-                                                 std::int32_t &x,
-                                                 std::int32_t &y) {
+                                           std::int32_t &x,
+                                           std::int32_t &y) {
   std::int32_t rel_x = 0;
   std::int32_t rel_y = 0;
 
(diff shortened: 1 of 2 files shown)


salvprest:
sdl is indeed trying to set a windowID, but for some reason is not able to, perhaps for the same reason that it is not able to normalize the coordinates of tfinger

morphis:
@brunocalmon There are no written timelines for when a MR gets merged.

src/anbox/platform/sdl/platform.cpp
```diff
@@ -216,7 +216,7 @@ void Platform::process_input_event(const SDL_Event &event) {
       if (!single_window_) {
         // As we get only absolute coordindates relative to our window we have to
         // calculate the correct position based on the current focused window
-        window = SDL_GetWindowFromID(event.window.windowID);
+        window = SDL_GetWindowFromID(focused_sdl_window_id);
```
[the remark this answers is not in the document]
salvprest: sure we can use event.motion.windowID here, I have putted focused_sdl_window_id here too, to have a generalized way for grab the window on the process_input_event() function, if you want I can revert this

src/anbox/platform/sdl/platform.cpp
```diff
@@ -266,14 +266,20 @@ void Platform::process_input_event(const SDL_Event &event) {
...
       touch_events.push_back({EV_KEY, BTN_TOOL_FINGER, 1});
 
-      window = SDL_GetWindowFromID(event.window.windowID);
-      if (window) {
-        SDL_GetWindowSize(window, &rel_x, &rel_y);
-        rel_x *= event.tfinger.x;
-        rel_y *= event.tfinger.y;
+      window = SDL_GetWindowFromID(focused_sdl_window_id);
+      // before SDL 2.0.7 on X11 tfinger coordinates are not normalized
+      if (!SDL_VERSION_ATLEAST(2,0,7) && (event.tfinger.x>1 || event.tfinger.y>1)) {
```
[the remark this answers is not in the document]
salvprest: ok

src/anbox/platform/sdl/platform.cpp
```diff
@@ -314,14 +320,20 @@ void Platform::process_input_event(const SDL_Event &event) {
 	  case SDL_FINGERMOTION: {
       touch_events.push_back({EV_ABS, ABS_MT_SLOT, 0});
 
-      window = SDL_GetWindowFromID(event.window.windowID);
-      if (window) {
-        SDL_GetWindowSize(window, &rel_x, &rel_y);
-        rel_x *= event.tfinger.x;
-        rel_y *= event.tfinger.y;
+      window = SDL_GetWindowFromID(focused_sdl_window_id);
+      // before SDL 2.0.7 on X11 tfinger coordinates are not normalized
```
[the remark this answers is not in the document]
salvprest: ok

src/anbox/platform/sdl/platform.cpp
```diff
@@ -370,6 +315,49 @@ void Platform::process_input_event(const SDL_Event &event) {
 #endif
 }
 
+bool Platform::calculate_touch_coordinates(const SDL_Event &event,
+                                                 std::int32_t &x,
```
morphis: Indentation is incorrect here

src/anbox/platform/sdl/platform.cpp
```diff
@@ -370,6 +315,49 @@ void Platform::process_input_event(const SDL_Event &event) {
...
+                                                 std::int32_t &y) {
+  std::int32_t rel_x = 0;
+  std::int32_t rel_y = 0;
+
+  SDL_Window *window = nullptr;
+
+  window = SDL_GetWindowFromID(focused_sdl_window_id_);
```
morphis: It's unfortunate that we can't get the window id from the event .. however if this approach works well lets go this way.

src/anbox/platform/sdl/platform.cpp
salvprest: I have used window_moved() and window_resized() as models, but now I see that in other file the convention seem different, I'll fix this

src/anbox/platform/sdl/platform.cpp
salvprest: yes it's unfortunate, the SDL team see the touch device as something more general of a touchscreen, so they don't want to set the windowID through input touch.
on my systems this approach work really well

ZZZXXX0110:
Any update lately?
Looks like we already have a working solution?

Hedronmx:
Can't wait for this to merge. Will change my linux experience.

ZZZXXX0110:
Hooray!!!  Now hopefully it won't take too long for the snap release to get updated!

PR #1045, opened 2019-02-03
Add Intel Core 2 Duo E7500 to the whitelist

Encritary:
It does support SSE 4.2 instructions, I've checked that.


Encritary:
@morphis oh, it seems like the site I got the info from had a mistake. But then how can I launch anbox on my E7500?

morphis:
@Encritary I am sorry, but without support for SSE 4.2 you can't unless somebody contributes changes to make Android build without it.

Issue #25, opened 2017-04-12
Touchscreen events don't work


intirix:
Make sure you are running the latest version of Anbox before reporting an issue.

**Anbox release (`anbox version`) and system info (`anbox system-info`):**
anbox 0.1.0

version: 0.1.0
os:
  name: Ubuntu Core
  version: 16
  snap-based: true
kernel:
  version: Linux version 4.4.0-72-generic (buildd@lcy01-17) (gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4) ) #93-Ubuntu SMP Fri Mar 31 14:07:41 UTC 2017
  binder: true
  ashmem: true
graphics:
  egl:
    vendor: Mesa Project
    version: 1.4 (DRI2)
    extensions:
      - EGL_CHROMIUM_sync_control
      - EGL_EXT_create_context_robustness
      - EGL_EXT_image_dma_buf_import
      - EGL_KHR_create_context
      - EGL_KHR_fence_sync
      - EGL_KHR_get_all_proc_addresses
      - EGL_KHR_gl_renderbuffer_image
      - EGL_KHR_gl_texture_2D_image
      - EGL_KHR_gl_texture_cubemap_image
      - EGL_KHR_image
      - EGL_KHR_image_base
      - EGL_KHR_image_pixmap
      - EGL_KHR_reusable_sync
      - EGL_KHR_surfaceless_context
      - EGL_KHR_wait_sync
      - EGL_MESA_configless_context
      - EGL_MESA_drm_image
      - EGL_MESA_image_dma_buf_export
      - EGL_NOK_swap_region
      - EGL_NOK_texture_from_pixmap
      - EGL_NV_post_sub_buffer
      - EGL_WL_bind_wayland_display
  gles2:
    vendor: Intel Open Source Technology Center
    vendor: OpenGL ES-CM 1.1 Mesa 12.0.6
    extensions:
      - GL_EXT_blend_minmax
      - GL_EXT_multi_draw_arrays
      - GL_EXT_texture_filter_anisotropic
      - GL_EXT_texture_lod_bias
      - GL_OES_byte_coordinates
      - GL_OES_fixed_point
      - GL_OES_stencil_wrap
      - GL_OES_compressed_paletted_texture
      - GL_OES_query_matrix
      - GL_OES_read_format
      - GL_OES_single_precision
      - GL_EXT_texture_compression_dxt1
      - GL_OES_draw_texture
[...]

PR #1073, opened 2019-03-17
touch always enabled and touch emulation

salvprest:
1. The first commit remove the ENABLE_TOUCH_INPUT flag, and enable always the touch driver, the problems connected with the touch support were caused by an incomplete implementation of the touch input handling, and now they are fixed, so we are limiting only functionalities.
2. With the second commit we are converting mouse inputs into touch inputs on the android side, android apps expect a touch screen, some of them work only with touch inputs, in this way we support a wider range of apps without losing any functionalities.
3. The third commit introduce the command no-touch-emulation, which disable the touch emulation, this command is thinked for developers that want to see how their apps work with a mouse


Fmstrat:
Since the initial comment doesn't mention it, is this PR intended to fix: https://github.com/anbox/anbox/issues/25

Just curious as many are tracking that issue.

salvprest:
> Since the initial comment doesn't mention it, is this PR intended to fix: #25
> 
> Just curious as many are tracking that issue.

my previous pr that was merged should have fixed the touch screen on linux, but the touch support is disabled by default, so for now you haven't seen any change, but the first commit that enable touch by default should fix #25 
but this should fix alot of other bugs related to apps that works only with touch inputs

Fmstrat:
> 
> 
> > Since the initial comment doesn't mention it, is this PR intended to fix: #25
> > Just curious as many are tracking that issue.
> 
> my previous pr that has been merged, should have fixed the touch screen on linux, but the touch support is disabled by default, so for now you haven't seen any change, but the first commit that enable touch by default should fix #25
> but this should fix alot of other bugs related to apps that works only with touch inputs

Thanks for the additional info. So once this is merged in and released, should touch be enabled for anbox in the edge branch? I understand the first PR implemented it, but without compiling manually I don't think there is a way to enable it, right?

salvprest:
exactly, for now you should compile anbox with the option ENABLE_TOUCH_INPUT enabled

Fmstrat:
@salvprest Sorry for the potentially dumb question, but I change that setting in `build/CMakeCache.txt` after `cmake ..` right?

salvprest:
you can edit the `build/CMakeCache.txt` file or run cmake with the command `cmake -DENABLE_TOUCH_INPUT=true`.
Or you can try to compile anbox with this pr by downloading it from my [repo](https://github.com/salvprest/anbox/tree/touch_emulation) so you don't have to change the cmake options

Fmstrat:
@salvprest Thanks for the help! Unfortunately the build works but running into an issue with the snap build to get it moving fully. https://github.com/anbox/anbox/issues/1076

Fmstrat:
Hmm.. I just built from your fork (had to use xenial), however touch is not working. Is there a specific branch I should use?

hdk5:
Fixed my shipgirls issue, thanks!

ZZZXXX0110:
Hope this can be merged soon.

awaken1ng:
Mouse wheel event is causing problems in [gun girls](https://play.google.com/store/apps/details?id=com.sunborn.girlsfrontline.en) (don't really have anything else to test on)

For example, in formation menu:
- Before scrolling:
Before scrolling
- After scrolling at least once:
After scrolling

As a side note, the game doesn't need touch emulation and works with mouse just fine

Messed around with the event for a bit, couldn't really find a solution (aside from ignoring it/disabling pointer entirely when touch emulation is enabled to avoid this), so reporting this just in case

salvprest:
thanks @awaken1ng, so this problem is caused by this pr? if you start anbox session manager with `anbox session-manager --no-touch-emulation` you have the same problem?

awaken1ng:
Works as expected if touch emulation is disabled
Also works fine with it enabled, until what it seems like any pointer event (which is mouse wheel scroll in this case), then it starts being weird

salvprest:
ok I'll add a commit to fix this

salvprest:
@awaken1ng I have tried this game with a real android device, and a real mouse, and this problem occur even there, when I use the mouse and then the touch, or vice versa, then that strange behavior occur, the only way of resolving the issue is restart the app, disconnect the mouse, or move the cursor in a neutral corner doesn't solve anything.
this app seem to use unity game engine, but other unity games that I have tried don't seem to have this problem.
so this is a problem related to the app.
A way to avoid this is implement an emulated mouse scroll with only touch events, but that kind of implementation have some drawbacks..

awaken1ng:
I see, so it's game specific, thank you for bothering to test it on real device
Yeah, emulating swipe-to-scroll with touch events can be very clunky, at this point maybe it would be better to have pinch-to-zoom mode and drop mouse wheel entirely

hdk5:
https://gfycat.com/keengloomyelkhound  
That's how Nox emulator handles mouse wheel events.  
Just scroll for swipe, hold Ctrl for pinch.  The more you scroll the longer is trail.  
It would be cool to have something similar.

salvprest:
@hdk5 thats seem cool, but with that kind of scroll, made with touch inputs, if you scroll above something that is draggable, probably you end with an unwanted object drag, I think

tuxflo:
Hey there, could you provide some more information about how to use this feature?
I've cloned the repo with the patches and built a snap of that and installed it.
Now I'm able to run `snap run anbox session-manager --no-touch-emulation` after that nothing happens. If I want to list the "Main Menu" I have to run `snap run anbox launch --package=org.anbox.appmgr --component=org.anbox.appmgr.AppViewActivity` However nothing touch related is changed, anbox still uses the mouse like when installing the "regular" snap without the patches. I'm trying to use a simple pdf viewer app (mupdf from f droid) working with pinch to zoom gestures.

salvprest:
@tuxflo this simply translate the mouse inputs in touch inputs, it don't implement gestures like the pinch to zoom

tuxflo:
@salvprest Ok then I totally misunderstood the name "no touch emulation" I thought the flag does NOT translate the input and directly passes touch events or even the complete device. But thanks anyway for the quick reply! I want to try some modifications on the code on my own, could you provide some tips how to run a self compiled local anbox build without building the snap package first?

src/anbox/cmds/session_manager.cpp
```diff
@@ -157,7 +160,8 @@ anbox::cmds::SessionManager::SessionManager()
     auto platform = platform::create(utils::get_env_value("ANBOX_PLATFORM", "sdl"),
                                      input_manager,
                                      display_frame,
-                                     single_window_);
+                                     single_window_,
+                                     no_touch_emulation_);
```
morphis: We should create a platform::Configuration struct and pass that instead here to avoid having too many arguments. The Configuration struct can take display_frame, single_window_ and no_touch_emulation_

salvprest: ok, having too many arguments was one of my concern

salvprest:
@tuxflo I build the snap package to try my modifications on multiple devices, so sorry I can't help you on this

present86:
> @tuxflo I build the snap package to try my modifications on multiple devices, so sorry I can't help you on this

did you manage to build the snap package? can't wait to try

src/anbox/cmds/session_manager.cpp
```diff
@@ -112,6 +112,9 @@ anbox::cmds::SessionManager::SessionManager()
   flag(cli::make_flag(cli::Name{"software-rendering"},
                       cli::Description{"Use software rendering instead of hardware accelerated GL rendering"},
                       use_software_rendering_));
+  flag(cli::make_flag(cli::Name{"no-touch-emulation"},
```
morphis: We should add a snap config item for this so users can easily run a `snap set anbox touch-emulation.enable=false`

salvprest: ok I don't know very well the snap package system, but I think that you mean something like what we do with the software-rendering.enable, that in snap-wrapper.sh is evaluated and then export a variable that is evaluated in the session-manager.cpp.
so ok adding something like this seem simple and can be useful

morphis: Correct, thanks for adding it

tuxflo:
@present86 I used the command `snapcraft cleanbuild` to create a custom snap package. If you are interested in multitouch as well you might take a look at the PR https://github.com/anbox/anbox/pull/1073 which sounds promising.

PR #1084, opened 2019-03-23
emugl: rebuild generated sources only when needed

mikma:
Some improvements to the build system which for example allows you to rerun make after a successful build without causing rebuild of the generated sources and libraries and programs that use them.


morphis:
LGTM, thanks!

PR #1086, opened 2019-03-24
Vpn fixes

mikma:
This branch contains fixes that are required for (almost) working VPN. The android.img also needs to be recreated with the modified scripts/anbox-init.sh, otherwise you need to manually run the following script on each boot.

```
#!/bin/sh

if [ -e "/dev/tun" ] ; then
	chown system:vpn /dev/tun
	chmod 0660 /dev/tun
fi
```

There is still one known issue, VpnService.protect() doesn't work yet. It means you won't be able to use a VPN service if the remote VPN gateway is covered by the VPN routes. I.e. you can't use a VPN service which inserts a default route.

### DIFF
--- a/scripts/anbox-init.sh
+++ b/scripts/anbox-init.sh
@@ -25,6 +25,11 @@ function prepare_filesystem() {
 		chown system:system /dev/$f
 		chmod 0666 /dev/$f
 	done
+
+	if [ -e "/dev/tun" ] ; then
+		chown system:vpn /dev/tun
+		chmod 0660 /dev/tun
+	fi
 }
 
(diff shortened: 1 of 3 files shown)


hotkopi:
> There is still one known issue, VpnService.protect() doesn't work yet. It means you won't be able to use a VPN service if the remote VPN gateway is covered by the VPN routes. I.e. you can't use a VPN service which inserts a default route.

Could you help explain more on the VpnService.protect() issue?

Does that mean those VPNs that provide anonymity won't work (since they require setting up default route)? Any quick/ manual fix possible to resolve the issue?

Thanks!

PR #1106, opened 2019-04-18
do not use constants for glFormat and glType

p-an:
Hi,
I am not absolutely sure about this, but the constants cause discrapancies in buffer sizes, which makes osmand (https://f-droid.org/en/packages/net.osmand.plus/) crash at start. With this change osmand runs correctly.

Pavel


morphis:
LGTM, thanks!

PR #1123, opened 2019-04-29
use GREATER and EQUAL instead of GREATER_EQUAL

salvprest:
GREATER_EQUAL require cmake 3.7 but I think that our cmake target is lower, on ubuntu 16.04 we have cmake 3.5.1
the automatic builds of the newer commits on snapcraft are failing for this reason i believe


morphis:
LGTM, thanks!

PR #1137, opened 2019-05-11
repair build on GCC 8.3.0

Logout22:
Use pre-defined macros to fill the systemd structures,
whose uninitialized field might otherwise
cause undefined behavior, as stated in the systemd header.

### DIFF
--- a/src/anbox/dbus/sd_bus_helpers.h
+++ b/src/anbox/dbus/sd_bus_helpers.h
@@ -27,91 +27,53 @@ namespace sdbus {
 namespace vtable {
 constexpr sd_bus_vtable start(uint64_t flags)
 {
-  sd_bus_vtable v{};
-  v.type = _SD_BUS_VTABLE_START;
-  v.flags = flags;
-  v.x.start = decltype(v.x.start){sizeof(sd_bus_vtable)};
-
-  return v;
+  return SD_BUS_VTABLE_START(flags);
 }
 
 constexpr sd_bus_vtable end() {
-  sd_bus_vtable v{};
-  v.type = _SD_BUS_VTABLE_END;
-
-  return v;
+  return SD_BUS_VTABLE_END;
 }
 
 constexpr sd_bus_vtable method_o(const char* member, const char* signature,
                                  const char* result,
                                  sd_bus_message_handler_t handler, size_t offset,
                                  uint64_t flags) {
-  sd_bus_vtable v{};
-  v.type = _SD_BUS_VTABLE_METHOD;
-  v.flags = flags;
-  v.x.method = decltype(v.x.method){member, signature, result, handler, offset};
-
-  return v;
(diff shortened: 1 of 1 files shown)


src/anbox/dbus/sd_bus_helpers.h
```diff
@@ -27,46 +27,28 @@ namespace sdbus {
...
-  sd_bus_vtable v{};
-  v.type = _SD_BUS_VTABLE_SIGNAL;
-  v.flags = flags;
-  v.x.signal = decltype(v.x.signal){member, signature};
-
-  return v;
+  return sd_bus_vtable SD_BUS_SIGNAL(member, signature, flags);
 }
 
 constexpr sd_bus_vtable property(const char* member, const char* signature,
```
razvancrainea: The same thing should be done for the `property` methods too; even if it works for now, it might break in the future.

Logout22: Sure. I just was not certain if you would like my fix or if it would get a green build.

razvancrainea: Well, I am not a maintainer, I've just hit the same bug and when I was about to submit a PR, I saw yours, and I thought it's worth the hint :).

Logout22: It is, thank you! I finished the fix accordingly.

morphis:
@Logout22 Sadly this breaks the snap build which still used the gcc version 5.3 from Ubuntu 16.04. Can you rework this to work with 5.3 too?

Logout22:
@morphis That will not be so easy. The problem is, as I just found out, that I am using the dbus initializers wrong. According to https://stackoverflow.com/a/25362712 , their syntax is not allowed in C++. So, as I cannot change the old system headers even if they would allow me (which is unlikely), I only have two options:

I can move the helpers' implementation into a C file and refer to them as external C functions in sd_bus_helpers.h, which would void the benefits of a constexpr. If the helpers are not on the critical path this may not be so bad, and I actually favour this solution as it avoids code duplication and ensures compatibility with the C++ standard and thereby other C++ compilers, e.g. clang.

Alternatively, I could put a C++-compatible copy of the sd_bus_vtable initializers into a separate file that I only include if the compiler version is too low. I would only go this way if the helpers are critical to anbox' performance. I do not like the idea of duplicating code in general, let alone code from system headers that may change for good reasons, normally without the application developer knowing or caring. Also, it would require more tedious work to implement.

Please tell me which solution you prefer, and I may eventually start to implement that one.

Thanks for raising awareness!

morphis:
@Logout22 Thanks for the detailed overview! I would like to go with option 1. Is that fine for you?

Logout22:
@morphis Oh, I just realized I forgot to answer. That was very impolite. Sorry!

Yes, of course, the pull request is open ( #1209 ). Thanks for your support!

PR #1201, opened 2019-07-22
Revert "repair build on GCC 8.3.0"

morphis:
Reverts anbox/anbox#1137

Breaks with gcc 5.3 on Ubuntu 16.04 which is our primary delivery channel via snaps.


morphis:
@Logout22 FYI

PR #1209, opened 2019-08-02
fix(anbox#1153): compile dbus helpers in C

Logout22:
This fixes the missing-field-initializers warning
and maintains compatibility with the C++ standard
(which pull request #1137 did not), thus working
also with older compilers that lack C struct initialization.


morphis:
@Logout22 Thanks!

PR #1352, opened 2020-02-02
platform/sdl: improve user experience when resize window

pingplug:
it is annoying when you want to drag something near the edge, and then the window is resized.

- do not resize or drag when window is fullscreen
- do not resize when window is maximized
- do not check button area in single window mode
- use thinner drag area in single window mode

set `window_resize_border` to 10 or less might be better


morphis:
LGTM, thanks!

PR #1358, opened 2020-02-09
musl fixes

necessarily-equal:
These changes are needed to compile Anbox with the musl libc. I heard there was an interest in upstreaming patches, so here's some!


morphis:
LGTM, thanks!

Issue #1357, opened 2020-02-09
Make Anbox build on not-x86_64 on postmarketOS (update the bundled cpu_features version?)


necessarily-equal:
Building Anbox on postmarketOS/Alpine is now possible (hooray!) but it saddly fails for non-x86_64 architectures, even ones that should work like AArch64. It fails with errors such as this one: https://paste.gnugen.ch/raw/HVvE. It would be great to get this fixed before merging, as to run Android apps on actual phones :P

6. ** Please paste the result of `anbox system-info` below:**

(Who knows, maybe that might be useful)

<details>
<summary>anbox system-info output</summary>
```
version: local-bd35e003cc
cpu:
  arch:  x86
  brand: Intel(R) Core(TM) i5-5300U CPU @ 2.30GHz
  features: 
    - aes
    - sse4_1
    - sse4_2
    - avx
    - avx2
os:
  name: Alpine Linux
  version: 
  snap-based: false
kernel:
  version: Linux version 5.4.4-0-lts (antoine@thinkpad-x-ccl) (gcc version 9.2.0 (Alpine 9.2.0)) #1-Alpine SMP Thu, 26 Dec 2019 22:37:35 UTC
  binder: true
  ashmem: true
graphics:
  egl:
    vendor: Mesa Project
    version: 1.4
    extensions:
      - EGL_ANDROID_blob_cache
      - EGL_ANDROID_native_fence_sync
      - EGL_CHROMIUM_sync_control
      - EGL_EXT_buffer_age
      - EGL_EXT_create_context_robustness
      - EGL_EXT_image_dma_buf_import
      - EGL_EXT_image_dma_buf_import_modifiers
      - EGL_IMG_context_priority
      - EGL_KHR_config_attribs
      - EGL_KHR_create_context
      - EGL_KHR_create_context_no_error
      - EGL_KHR_fence_sync
      - EGL_KHR_get_all_proc_addresses
      - EGL_KHR_gl_colorspace
      - EGL_KHR_gl_renderbuffer_image
      - EGL_KHR_gl_texture_2D_image
      - EGL_KHR_gl_texture_3D_image
      - EGL_KHR_gl_texture_cubemap_image
      - EGL_KHR_image
      - EGL_KHR_image_base
      - EGL_KHR_image_pixmap
      - EGL_KHR_no_config_context
      - EGL_KHR_reusable_sync
      - EGL_KHR_surfaceless_context
      - EGL_EXT_pixel_format_float
[...]

PR #1362, opened 2020-02-17
fix arm build: avoid including x86-only header from cpu_features on arm

necessarily-equal:
It breaks build on Alpine/postmarketOS, and is explicitely an error on more recent version of cpu_features: https://github.com/google/cpu_features/blob/master/include/cpuinfo_x86.h#L207.

protecting the include with `#ifdef` fix the problem, as every use of cpu_features is already protected.

### DIFF
--- a/src/anbox/cmds/check_features.cpp
+++ b/src/anbox/cmds/check_features.cpp
@@ -19,7 +19,9 @@
 #include "anbox/utils.h"
 
 #include "cpu_features_macros.h"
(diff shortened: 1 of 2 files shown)


necessarily-equal:
Merging this along as updating the cpu_feature version bundled (see https://github.com/anbox/anbox/issues/1357) should fix the build totally.

morphis:
LGTM, thanks!

PR #1383, opened 2020-03-10
added latest cpu_features

sreehax:
This is to get an armhf/armv7/aarch64 build working in anbox with no modifications, more specifically for postmarketOS. see #1357


morphis:
This fails still on CI, see https://travis-ci.org/github/anbox/anbox/builds/660876134

```
-- Checking for module 'gtest'
--   No package 'gtest' found
CMake Error at /usr/share/cmake-3.10/Modules/FindPkgConfig.cmake:415 (message):
  A required package was not found
Call Stack (most recent call first):
  /usr/share/cmake-3.10/Modules/FindPkgConfig.cmake:593 (_pkg_check_modules_internal)
  cmake/FindGMock.cmake:1 (pkg_check_modules)
  CMakeLists.txt:120 (find_package)
```

Can you get this fixed?

PR #1404, opened 2020-03-31
Added missing dependency

TheOneWithTheBraid:
Added python2 as build dependency. On modern system it no longer is preinstalled by default.


morphis:
Thanks!

lucasfcnunes:
python2 apt install changed #1451

PR #1415, opened 2020-04-12
session-manager: implement configurable server-side decoration

### LINKED ISSUE #1308: Why anbox doesn't have a normal windows bar  ?
Hi, I just wanted to know why anbox didn't have a normal title bar with resize, minimise etc from the DE ? Is this by design ? Why ? Also the border to resize the window are too much bigger I can't click on certain button sometimes and its really frustrating

thdaemon:
Add command line option `--server-side-decoration` and `ANBOX_FORCE_SERVER_SIDE_DECORATION` env-var for session-manager to use server-side decoration. Both methods can activate it.

When using server-side decoration, the hit test handler will NOT be set beacuse we don't need to simulate the behavior of the title bar and resize area.

Currently android framework layer will render caption (client-side decoration), in order to let the android framework layer know that sometimes we need server-side decoration. I introduce a special system property `ro.anbox.no_decorations`. When it is true, hide the android client-side decoration caption unconditionally. When server-side decoration is enabled, this property is set to `1`.

Therefore, server-side decoratio needs to be used with another pull request to anbox's android framework, otherwise you will get two title bars. https://github.com/anbox/platform_frameworks_base/pull/6

Fix #1308

### DIFF
--- a/scripts/snap-wrapper.sh
+++ b/scripts/snap-wrapper.sh
@@ -47,6 +47,10 @@ if [ "$(snapctl get touch-emulation.enable)" = false ]; then
 	export ANBOX_ENABLE_TOUCH_EMULATION=false
 fi
 
+if [ "$(snapctl get server-side-decoration.enable)" = true ]; then
+	export ANBOX_FORCE_SERVER_SIDE_DECORATION=true
+fi
+
 # Use custom Anbox binary for debugging purposes if available
 ANBOX="$SNAP"/usr/bin/anbox
 if [ -e "$SNAP_COMMON"/anbox.debug ]; then
--- a/src/anbox/cmds/session_manager.cpp
+++ b/src/anbox/cmds/session_manager.cpp
@@ -115,6 +115,9 @@ anbox::cmds::SessionManager::SessionManager()
   flag(cli::make_flag(cli::Name{"no-touch-emulation"},
                       cli::Description{"Disable touch emulation applied on mouse inputs"},
                       no_touch_emulation_));
+  flag(cli::make_flag(cli::Name{"server-side-decoration"},
+                      cli::Description{"Prefer to use server-side decoration instead of client-side decoration"},
+                      server_side_decoration_));
 
   action([this](const cli::Command::Context &) {
     auto trap = core::posix::trap_signals_for_process(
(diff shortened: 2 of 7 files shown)


thdaemon:
@morphis 
Thank you very much for giving me the idea about configurable server-side decoration. But I am not sure whether the modification of the android framework code is appropriate. There is no problem with my test. If I break something, please let me know.

thdaemon:
If you want to test but do not want to build android image yourself. I have built one and uploaded to here: https://github.com/thdaemon/anbox/releases/tag/ssd-pr-v1
If you want to build yourself, using anbox/platform_frameworks_base#6
You can use the image with this pull request and test configurable server-side decoration

cyberl33t:
No, it is NOT working. Exported ANBOX_FORCE_SERVER_SIDE_DECORATION=true in my .xinitrc, rebuilt anbox from git just 15 mintes ago (via Arch AUR), rebooted, and the csd bar is STILL there!!! What's wrong?

thdaemon:
> No, it is NOT working. Exported ANBOX_FORCE_SERVER_SIDE_DECORATION=true in my .xinitrc, rebuilt anbox from git just 15 minutes ago (via Arch AUR), rebooted, and the csd bar is STILL there!!! What's wrong?

You MUST use a patched Android image. My built is here: https://github.com/thdaemon/anbox/releases/tag/ssd-pr-v1
Changes to anbox upstream Android repos have been merged too. Maybe next offical Android build will included CSD supports
Thanks.

alxlg:
@thdaemon when will the Android image be updated? I tried to use Anbox with your custom image following the instructions in the official repo but it never worked for me. Having this patch in the Android image distributed by Anbox would be much better.

thdaemon:
> @thdaemon when will the Android image be updated? I tried to use Anbox with your custom image following the instructions in the official repo but it never worked for me. Having this patch in the Android image distributed by Anbox would be much better.

It's up to morphis to decide when to build the anbox official Android image. i just wrote the patch and am not a member of the anbox group.
The SSD should work when used with my Android image build. Maybe because anbox doesn't use this custom image. I'm sorry that I'm not sure how to set custom image for the snap edition of anbox, it seems to handle android.img downloads automatically. If you're using anbox built directly from source, you can just clean up the `/var/lib/anbox/` directory (this will delete user data on your anbox Android, so be sure to back it up), and then put android.img in this directory.

alxlg:
@thdaemon cc @morphis for updating the Android image

FYI I followed the third paragraph of this part of the documentation: https://github.com/anbox/anbox/blob/master/docs/build-android.md 

...but Anbox service doesn't start with custom Android image (no problem with the built-in one). Sorry I can't check my PC for more details on the error now.

alxlg:
The Android image still need to be updated, right?

Issue #1382, opened 2020-03-09
Debian make fails with "no rule to make target 'gmock/libgmock.a"


Aebian:
Only issue I found here on the tracker was for arch and not directly related to my issue.
### sudo ldconfig -v | grep gtest
```
libgtest.so -> libgtest.so
  libgtest_main.so -> libgtest_main.so
ldconfig: /lib/x86_64-linux-gnu/ld-2.29.so is the dynamic linker, ignoring
```

### sudo ldconfig -v | grep gmock
```
libgmock_main.so -> libgmock_main.so
ldconfig: /lib/x86_64-linux-gnu/ld-2.29.so is the dynamic linker, ignoring
        libgmock.so -> libgmock.so
```

### Extact error is: 

```
[  3%] Built target GMock
[ 10%] Built target process-cpp
[ 13%] Built target emugen
[ 15%] Built target OpenglCodecCommon
[ 16%] Built target renderControl_dec
[ 18%] Built target GLESv2_dec
[ 20%] Built target GLESv1_dec
[ 25%] Built target emugl_common
[ 27%] Built target OpenGLESDispatch
[ 29%] Built target xdg
[ 30%] Built target xdg_test
[ 31%] Built target backward
[ 31%] Built target backward_object
[ 36%] Built target cpu_features
[ 37%] Built target list_cpu_features
[ 40%] Built target anbox-protobuf
[ 89%] Built target anbox-core
[ 90%] Built target anbox
make[2]: *** No rule to make target 'gmock/libgmock.a', needed by 'tests/anbox/android/intent_tests'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:1919: tests/anbox/android/CMakeFiles/intent_tests.dir/all] Error 2
make: *** [Makefile:141: all] Error 2

```

### Steps I did after all requirements where installed / built
1.) git clone <repo-url>
2.) cd anbox
3.) mkdir build 
4.) cd build
5.) cmake ..
6.) sudo make install

I have built gtest and gmock from source 

### How I built googletest
```
git clone https://github.com/google/googletest.git
cd googletest
mkdir build && cd build
cmake .. -DBUILD_SHARED_LIBS=ON -DINSTALL_GTEST=ON -DCMAKE_INSTALL_PREFIX:PATH=/usr
make -j8
sudo make install
sudo ldconfig
```
### Second Confirmation that gtest framework is installed:
```
→ pkg-config --libs --cflags gtest
[...]

PR #1423, opened 2020-04-16
use gtest in a sane way when it is packaged properly.

necessarily-equal:
some distributions, at least Arch and Alpine, ship a binary package with
a pkg-config. On those platforms, tests are disabled or such a patch is applied.

I tested this to still build on Debian and Alpine. Arch should be unaffected.


necessarily-equal:
also applies when google-test is built from source: https://github.com/anbox/anbox/issues/1382

PR #1426, opened 2020-04-17
container-manager: support setting --force-squashfuse

AkihiroSuda:
`anbox container-manager --force-squashfuse` enforces to use `squashfuse`.

This flag is useful for avoiding `anbox[70]: Could not create loopback device: No such file or directory` error that happens when running Anbox inside Docker : https://github.com/aind-containers/aind/issues/5

### DIFF
--- a/src/anbox/cmds/container_manager.cpp
+++ b/src/anbox/cmds/container_manager.cpp
@@ -55,6 +55,9 @@ anbox::cmds::ContainerManager::ContainerManager()
   flag(cli::make_flag(cli::Name{"use-rootfs-overlay"},
                       cli::Description{"Use an overlay for the Android rootfs"},
                       enable_rootfs_overlay_));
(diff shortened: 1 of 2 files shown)


src/anbox/cmds/container_manager.cpp
```diff
@@ -55,6 +55,9 @@ anbox::cmds::ContainerManager::ContainerManager()
   flag(cli::make_flag(cli::Name{"use-rootfs-overlay"},
                       cli::Description{"Use an overlay for the Android rootfs"},
                       enable_rootfs_overlay_));
+  flag(cli::make_flag(cli::Name{"use-squashfuse"},
```
morphis, line 58: Can we call this "--force-squashfuse" instead?

AkihiroSuda: updated

PR #1429, opened 2020-04-18
WIP: scripts: make image creation time and modification times reproducibles

necessarily-equal:
see
https://reproducible-builds.org/docs/source-date-epoch/
and
https://reproducible-builds.org/docs/archives/

There are built-in features in mksquashfs 4.4 to do just that,
but this allows using the previous version which is still widespread
(Ubuntu ships it starting with 19.10, so there are no LTS with 4.4)

This has been tested to work with Ubuntu 16.04 LTS.


necessarily-equal:
ToDo: investigate BUILD_DATETIME.

necessarily-equal:
pop

PR #1454, opened 2020-05-14
armv7 tweaks

necessarily-equal:
This backport two fixes by @Minecrell used in the postmarketOS Anbox image for armv7.

The first make the Anbox image work on recent kernels, as 32bits binder support was dropped. The second is for consistency.


necessarily-equal:
Someone on the Telegram channel had an issue about this, which is what reminded me I should make a MR.

PR #1465, opened 2020-05-20
Include glm via system headers

mwkroening:
This removes the in-tree `glm` source and instead uses `glm` from the system libraries.

`glm` was already installed in the Docker container. I added `glm` as build dependency to `README.md` and `snapcraft.yaml`.

What do you think?


mwkroening:
I am experiencing a rendering issue with this (black windows). I'll have to investigate further.

mwkroening:
I added GLM 0.9.9.0 compatibility. Now everything is working fine.

PR #1480, opened 2020-06-02
rpc: use ByteSizeLong from protobuf

necessarily-equal:
ByteSize() is deperecated, and BiteSizeLong() returns a size_t already,
which is what we use. This way, we also avoid potential overflows when
casting to int back and from.


necessarily-equal:
I see that CI doesn't like it. I guess it uses an older version of protobuf.

necessarily-equal:
CI is fixed now. I believe this is ready for merging

necessarily-equal:
still ready

Fuseteam:
@morphis can this be reviewed?

necessarily-equal:
still ready

PR #1490, opened 2020-06-08
Build snap in CI instead of just Anbox

morphis:


necessarily-equal:
Hi, how about building both with and without snap?

morphis:
@necessarily-equal What would we win when doing that?

necessarily-equal:
Testing dofferent versions of Ubuntu and ensuring the CI still builds on non-snap? (I don't think anything should break on a regular build if the snap builds, but then I also wouldn't have thought the snap build couldn't break if it worked outside)

PR #1309, opened 2019-12-15
many: add support for binderfs

morphis:
binderfs is a new in kernel filesystem which abstracts dynamic
allocation of binder devices and makes it suitable for in container
use. See https://www.kernel.org/doc/html/latest/admin-guide/binderfs.html
and https://brauner.github.io/2019/01/09/android-binderfs.html for more
details.

We use binderfs when available (starting with Ubuntu 19.04 and its
standard kernel) and otherwise fallback to the static /dev/binder if
available.

Fixes #1273

Issue #1240, opened 2019-09-20
Implement support for binderfs instead of hardcoded paths


KOLANICH:
In Ubuntu Disco new version of binder module is used that uses binderfs instead of `/dev/binder`. The benefit and drawback is that it has to be explicitly mounted and then the files created with some ioctls. This benefits isolation. But anbox currently only has the hardcoded files paths. This should be fixed.

KOLANICH:
stalebot and lockbot are shit and shouldn't be used

mrvik:
This is a real issue when using the linux builtin binderfs support.
A temporary workaround is to run (as root):

```
mount -t binder binder /dev/binderfs #Or the location you prefer
ln -s /dev/binderfs/binder /dev/ #Change /dev/binderfs to the location specified above
```
Anyway, this is a ugly hack

paulvi:
#20 
#1309

Issue #20, opened 2017-04-11
upnp media renderer network discovery not working


monte-monte:
```
version: 0.1.0
os:
  name: Ubuntu Core
  version: 16
  snap-based: true
kernel:
  version: Linux version 4.4.0-72-generic (buildd@lcy01-17) (gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4) ) #93-Ubuntu SMP Fri Mar 31 14:07:41 UTC 2017
  binder: true
  ashmem: true
graphics:
  egl:
    vendor: Mesa Project
    version: 1.4 (DRI2)
    extensions:
      - EGL_CHROMIUM_sync_control
      - EGL_EXT_create_context_robustness
      - EGL_EXT_image_dma_buf_import
      - EGL_KHR_create_context
      - EGL_KHR_fence_sync
      - EGL_KHR_get_all_proc_addresses
      - EGL_KHR_gl_renderbuffer_image
      - EGL_KHR_gl_texture_2D_image
      - EGL_KHR_gl_texture_cubemap_image
      - EGL_KHR_image
      - EGL_KHR_image_base
      - EGL_KHR_image_pixmap
      - EGL_KHR_reusable_sync
      - EGL_KHR_surfaceless_context
      - EGL_KHR_wait_sync
      - EGL_MESA_configless_context
      - EGL_MESA_drm_image
      - EGL_MESA_image_dma_buf_export
      - EGL_NOK_swap_region
      - EGL_NOK_texture_from_pixmap
      - EGL_NV_post_sub_buffer
      - EGL_WL_bind_wayland_display
  gles2:
    vendor: Intel Open Source Technology Center
    vendor: OpenGL ES-CM 1.1 Mesa 12.0.6
    extensions:
      - GL_EXT_blend_minmax
      - GL_EXT_multi_draw_arrays
      - GL_EXT_texture_filter_anisotropic
      - GL_EXT_texture_lod_bias
      - GL_OES_byte_coordinates
      - GL_OES_fixed_point
      - GL_OES_stencil_wrap
      - GL_OES_compressed_paletted_texture
      - GL_OES_query_matrix
      - GL_OES_read_format
      - GL_OES_single_precision
      - GL_EXT_texture_compression_dxt1
      - GL_OES_draw_texture
      - GL_OES_point_size_array
      - GL_OES_point_sprite
      - GL_EXT_texture_format_BGRA8888
[...]

PR #1520, opened 2020-06-30
Update docs for unable to access '/ dev / binder'

paulvi:
wording taken from https://github.com/anbox/anbox-modules/issues/20#issuecomment-570204411 @morphis comment
ref #20 #1309 #1240


morphis:
LGTM, thanks!

PR #1522, opened 2020-07-02
Allow setting sensors value via dbus calls (eg. temperature)

tomasz-grobelny:
This pull request allows users of anbox to set temperature of Goldfish temperature sensor via dbus call like this:
```
dbus-send --session --dest=org.anbox --print-reply /org/anbox org.freedesktop.DBus.Properties.Set string:org.anbox.Sensors string:Temperature variant:double:25.1
```
This also introduces two new libraries to simplify code: sdbus-cpp and boost::di. It requires a C++ compiler supporting C++17.

The code supports only temperature for now. After this gets merged the plan is to add support for other sensors and location.

Code is ready for review.


src/anbox/cmds/session_manager.cpp
```diff
@@ -222,10 +227,13 @@ anbox::cmds::SessionManager::SessionManager()
...
     // The qemu pipe is used as a very fast communication channel between guest
     // and host for things like the GLES emulation/translation, the RIL or ADB.
+    const auto injector = di::make_injector(
+        di::bind<Renderer>().to(gl_server->renderer()),
+        di::bind<Runtime>().to(rt));
     auto qemu_pipe_connector =
         std::make_shared<network::PublishedSocketConnector>(
             utils::string_format("%s/qemu_pipe", socket_path), rt,
-            std::make_shared<qemu::PipeConnectionCreator>(gl_server->renderer(), rt));
+            injector.create<std::shared_ptr<qemu::PipeConnectionCreator>>());
```
necessarily-equal, line 236: Hey, you may want to document that new dependency injection thing, and why qemu is affected. I don't find this clear. (I'm not used to dependency injection, mind you)

src/anbox/qemu/sensors_message_processor.cpp
```diff
@@ -16,28 +16,69 @@
...
-SensorsMessageProcessor::~SensorsMessageProcessor() {}
+void SensorsMessageProcessor::handle_command(const string &command) {
+  if (!(list_sensors(command) || handle_set(command) || handle_set_delay(command)))
+    ERROR("Unknown command: " + command);
+}
+
+bool SensorsMessageProcessor::handle_set_delay(const string &command) {
+  if (!boost::starts_with(command, "set-delay:"))
+    return false;
+  delay = boost::lexical_cast<int>(command.substr("set-delay:"s.length()));
```
necessarily-equal, line 58: to be honest, maybe I'm used to C, but I find this much more clear and readable:
```C
if (!sscanf(command, "set-delay:%i", &delay))
    error();
```

compare: "scan string "command", looking for "set-delay:" followed by an int, and put me the int in $delay."

and the alternative "delay is the result of the conversion to an int of the $command string but where you dropped the $n first int, where n is what you get when you take "set-delay" as a string and then count the characters"

src/anbox/qemu/sensors_message_processor.cpp
```diff
@@ -16,28 +16,69 @@
...
-  // We don't support sensors yet so we mark all as disabled
-  int mask = 0;
-  char buf[12];
-  snprintf(buf, sizeof(buf), "%d", mask);
-  send_header(strlen(buf));
-  messenger_->send(buf, strlen(buf));
+bool SensorsMessageProcessor::list_sensors(const string &command) {
+  if (command != "list-sensors")
+    return false;
+  int mask = 8;
```
necessarily-equal, line 72: I assume mask did change because the temperature sensor is now supported. How about documenting that instead of making it a magic number?

src/anbox/dbus/ApplicationManagerClient.h
```diff
@@ -0,0 +1,22 @@
+#include <sdbus-c++/sdbus-c++.h>
```
necessarily-equal, line 1: Please add a GPL-3.0 header

src/anbox/dbus/ApplicationManagerServer.cpp
```diff
@@ -0,0 +1,37 @@
+
```
necessarily-equal, line 1: ditto

src/anbox/dbus/ApplicationManagerServer.h
```diff
@@ -0,0 +1,23 @@
+#include <sdbus-c++/sdbus-c++.h>
```
necessarily-equal, line 1: ditto

src/anbox/dbus/ApplicationManager.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
necessarily-equal, line 1: IMO, a plain description of an interface shouldn't be copyrightable, but my opinions likely doesn't matter on legal terms. You may want to add some GPL header or a SPDX-License-Identifier.

src/anbox/dbus/Sensors.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
necessarily-equal, line 1: same question applies

src/anbox/qemu/sensors_message_processor.cpp
tomasz-grobelny: You might be right... Not sure how would sscanf work with `std::atomic` though. How about solution with Boost Sprit X3?

src/anbox/qemu/sensors_message_processor.cpp
tomasz-grobelny: Would enum be enough?

src/anbox/dbus/ApplicationManager.xml
tomasz-grobelny: Any idea how SPDX-License-Identifier would look like in XML document? I may add just a plain text comment though (should be something more permissive than GPL I think).

necessarily-equal: that's a good point.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!-- SPDX-License-Identifier: MIT -->
```

should do, but @morphis should also say if he's ok with MIT license for that. Note that it's also possible to set it to something like `(GPL-2.0 OR MIT)`.

src/anbox/qemu/sensors_message_processor.cpp
necessarily-equal: IMO yes

src/anbox/cmds/session_manager.cpp
necessarily-equal: I'm afraid I don't know well enough the code base for that. Maybe adding a comment before the `const auto injector` while moving the qemu comment down would do?

src/anbox/qemu/sensors_message_processor.cpp
necessarily-equal: My personal opinion on the question is that the code was actually more readable before. In 10 minutes, I could not grasp what x3 really is, or how the new code actually works (like, I get what it's supposed to do, but I'd be quite embarrassed if I had to debug it). I think if you've not used x3 or C++ in a professional setting, this version is harder to read.

src/anbox/dbus/SensorsServer.h
```diff
@@ -0,0 +1,39 @@
...
+      : sdbus::AdaptorInterfaces<org::anbox::Sensors_adaptor>(connection, std::move(objectPath)), impl_(impl) {
+    registerAdaptor();
+  }
+
+  virtual ~SensorsServer() {
+    unregisterAdaptor();
+  }
+
+  double Temperature() override;
+  void Temperature(const double& value) override;
```
morphis, line 35: See comments above

src/anbox/qemu/sensors_message_processor.h
```diff
@@ -18,23 +18,32 @@
...
   ~SensorsMessageProcessor();
 
  protected:
-  void handle_command(const std::string &command) override;
+  void handle_command(const std::string& command) override;
 
  private:
-  void list_sensors();
+  void send_message(const std::string& message);
+  std::shared_ptr<application::SensorsState> sensorsState_;
```
morphis, line 37: -> sensors_state_

src/anbox/qemu/sensors_message_processor.h
```diff
@@ -18,23 +18,32 @@
...
+  std::atomic<int> delay = 200;
+  std::atomic<bool> temperature = false;
+  std::atomic<bool> run_thread = true;
+  std::thread thread;
```
morphis, line 41: All variables should have a trailing `_` in their name

src/anbox/cmds/session_manager.cpp
```diff
@@ -221,10 +227,13 @@ anbox::cmds::SessionManager::SessionManager()
 
     // The qemu pipe is used as a very fast communication channel between guest
     // and host for things like the GLES emulation/translation, the RIL or ADB.
+    const auto injector = di::make_injector(
```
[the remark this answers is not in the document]
tomasz-grobelny: No, it is not required. It is another proposal to use in anbox, totally independent of sdbus-cpp. The purpose is to limit code that is not necessarily very productive (passing arguments and instantiating objects) while still keeping the code explicit (object has to list its dependencies). The library should be used in a number of places in anbox besides those that I used it in.

Are you against this approach in general or is it more about cost to benefit ratio?

src/anbox/dbus/ApplicationManagerServer.cpp
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+#include "anbox/dbus/ApplicationManagerServer.h"
+
+#include <sstream>
+
+#include "anbox/android/intent.h"
+#include "anbox/logger.h"
+
+void ApplicationManagerServer::Launch(const std::map<std::string, sdbus::Variant>& intentDict, const std::string& stack) {
```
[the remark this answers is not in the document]
tomasz-grobelny: Here it is a bit more tricky - this is the spelling generated from `ApplicationManager.xml` file. It had `Launch` method and I used this spelling to keep the existing interface. Therefore we can keep the `Launch` casing for methods (and for consistency we would have `TryLaunch`) or we can change the interface of `ApplicationManager`. As I know `sdbus-cpp` does not the the option of changing case when generating code.

src/anbox/cmds/launch.cpp
```diff
@@ -219,7 +207,6 @@ anbox::cmds::Launch::Launch()
     // going to launch the real application now.
     ss.reset();
 
-    const auto success = try_launch_activity(app_mgr);
-    return success ? EXIT_SUCCESS : EXIT_FAILURE;
+    return client.TryLaunch(intent_, stack_) ? EXIT_SUCCESS : EXIT_FAILURE;
```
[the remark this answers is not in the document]
tomasz-grobelny: Why would that be when we are talking about code in `anbox::cmds::Launch::Launch()` method?

src/anbox/dbus/ApplicationManagerServer.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+    DEBUG("Launching %s", intent);
+    impl_->launch(intent, anbox::graphics::Rect::Invalid, launch_stack);
+  } catch (std::exception& err) {
+    ERROR("Failed to launch application: %s", err.what());
+    throw sdbus::Error("org.anbox.InternalError", err.what());
+  }
+}
+
+bool ApplicationManagerServer::Ready() {
```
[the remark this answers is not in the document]
tomasz-grobelny: See above - this comes directly from existing DBus interface.

src/anbox/cmds/launch.cpp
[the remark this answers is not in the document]
morphis: The convention for method names in the code is to use `try_launch` rather than `TryLaunch`. It's for consistency.

src/anbox/cmds/session_manager.cpp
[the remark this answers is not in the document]
morphis: I am not against it in general but it adds another out-of-tree dependency we have to care and I don't think the small benefit outweighs that.

src/anbox/dbus/ApplicationManagerServer.cpp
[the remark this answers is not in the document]
morphis: Ah I see. Then lets keep this as is.

tomasz-grobelny:
> 
> 
> I think we're mostly good. Can you as last change squash your git commits so we have only those doing real changes but not `Fix build, attempt 2` etc.?

I had a few busy days recently and there are a few equally busy ahead, will get back to that as soon as possible.

src/anbox/cmds/session_manager.cpp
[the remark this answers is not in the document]
tomasz-grobelny: Reverted for now. Maybe one day it will make sense to make another commit that reworks significant number of such cases so that it justifies the dependency on Boost::DI.

src/anbox/qemu/sensors_message_processor.h
tomasz-grobelny: Done.

tomasz-grobelny:
> 
> 
> > I think we're mostly good. Can you as last change squash your git commits so we have only those doing real changes but not `Fix build, attempt 2` etc.?
> 
> I had a few busy days recently and there are a few equally busy ahead, will get back to that as soon as possible.

Please have a look at another PR: https://github.com/anbox/anbox/pull/1540

PR #1525, opened 2020-07-07
snap: upgrade snap to 18.04

necessarily-equal:
This also modernize the snap a bit.

Note that this removes the snap desktop integration. I'm not sure if it was used by anything.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,11 +1,5 @@
 name: anbox
-version: 4
-version-script: |
-  if [ "$SNAPCRAFT_GRADE" = "stable" ]; then
-    echo $SNAPCRAFT_PROJECT_VERSION
-  else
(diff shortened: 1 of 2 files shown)


necessarily-equal:
Capture d’écran de 2020-07-08 01-10-23

Tested working

snap/snapcraft.yaml
```diff
@@ -1,18 +1,13 @@
 name: anbox
-version: 4
-version-script: |
-  if [ "$SNAPCRAFT_GRADE" = "stable" ]; then
-    echo $SNAPCRAFT_PROJECT_VERSION
-  else
-    echo $SNAPCRAFT_PROJECT_VERSION-$(git rev-parse --short HEAD)
-  fi
+version: git
```
morphis, line 2: Why are we dropping the actual version here? As there are no stable releases of Anbox we should fix what we had so far an use `4+gitr<git revision>`

PR #1545, opened 2020-07-31
Using C++-Style cast instead of C-Style cast.

Old-Nukkit:
It's good to use C++ cast, its safer.


mwkroening:
Since this affects an external project, I don't think that we should divert from upstream for changes not affecting functionality.

PR #1464, opened 2020-05-20
Include cpu_features via git submodule

mwkroening:
This replaces the in-tree copy of `cpu_features` with a git submodule updating it to the latest release of `v0.4.1`.

I am not sure, which version was being used in #1383 to resolve #1357, but this will make an update trivial, if necessary.

What do you think?

PR #1548, opened 2020-08-01
Include cpu_features via git submodule

mwkroening:
This is #1464 rebased on latest master.

I'll look into building the snap.


mwkroening:
So, I had complications with getting an Ubuntu VM up and running, but here we go!

The only thing I changed from #1464 aside from rebasing on latest master was changing the submodule url from ssh to https for systems like my VM, where ssh has not been configured.
I successfully built a snap with the Ubuntu 18.04 packaged snapcraft.

It does not look like @necessarily-equal had an issue with the url though. @necessarily-equal, could you try to reproduce the error? Perhaps dce1d1e2e14efa385af2b0baa6334678227fec08, which was not included in the last PR, fixed the issue.

morphis:
@mwkroening If we have the snap building fine with this, we should be good to go.

necessarily-equal:
hum, well, this time it seems to have worked, so I guess it's fine

PR #1568, opened 2020-08-17
20.04 snap

necessarily-equal:
This upgrades the snap to the 20.04 version. @tomasz-grobelny figured out the second part, cheers!

@morphis: before merging this, maybe we could fix some inconsistency. The `snapcraft.yaml` contains `grade: devel`, but also contains the `snapcraftctl set-grade stable` command. Which should we keep?

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -6,7 +6,7 @@ description: |
   separate the Android system fully from the host.
 confinement: devmode
 grade: devel
-base: core18
+base: core20
 architectures: [amd64]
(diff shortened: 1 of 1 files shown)


necessarily-equal:
Note that I didn't test the swiftshader commit

morphis:
@necessarily-equal We should keep it to `devel` as long as we can't get out of beta/edge in the snap store due to missing confinement.

morphis:
@necessarily-equal Going to merge this, can you do a follow up PR fixing that?

PR #1584, opened 2020-08-24
gitignore: drop outdated ignores

necessarily-equal:
kernel modules and debian files are not part of this repo anymore, so they don't need to have ignored files in gitignore.


morphis:
LGTM, thanks!

PR #1587, opened 2020-08-26
snap: cleanup devel/stable ambiguity

necessarily-equal:
the snap contains `grade: devel` at the top already


morphis:
LGTM, thanks!

Issue #1579, opened 2020-08-22
Installation fails on Ubuntu 20.04: "The repository does not have a Release file"


JonasVautherin:
Multiple issues already report this somehow, but I wanted to make it clearer.

Instructions say that Ubuntu 20.04 is a [supported distribution](https://github.com/anbox/anbox/blob/master/README.md#supported-linux-distributions):

```
At the moment we officially support the following Linux distributions:

Ubuntu 18.04 (bionic)
Ubuntu 20.04 (focal)
```

[Installation instructions](https://github.com/anbox/anbox/blob/master/docs/install.md#install-anbox) say that one should install Anbox through a PPA on Launchpad:

```
At the moment we only have packages prepared for Ubuntu in a PPA on Launchpad.
```

However, [the PPA command](https://github.com/anbox/anbox/blob/master/docs/install.md#install-necessary-kernel-modules) fails on Ubuntu 20.04:

```
$ sudo add-apt-repository ppa:morphis/anbox-support
[...]
E: The repository 'http://ppa.launchpad.net/morphis/anbox-support/ubuntu focal Release' does not have a Release file.
```

Something should be done to clarify this situation: either by fixing the PPA, or by giving a workaround in the installation instructions.

In the meantime, note that the failing PPA can be removed with:

```
$ sudo add-apt-repository --remove ppa:morphis/anbox-support
```

JonasVautherin:
Thanks @GizmoChicken! I managed to sign it on my Ubuntu 20.04. I opened a PR to update the documentation accordingly: https://github.com/anbox/anbox/pull/1588.

Issue #1570, opened 2020-08-18
If wanting to run Anbox on Ubuntu 20.04 with Secure Boot enabled (which is possible!), must use “kmodsign” for the “ashmem” module


GizmoChicken:
# Background #

In a [post on AskUbuntu](https://askubuntu.com/questions/1267823/module-ashmem-doesnt-load-with-secure-boot-on-but-binder-does-load-with-se), I reported that:

>I'm running **Ubuntu 20.04.1** with default **kernel 5.4.0-42-generic** (installed from Main) on a UEFI system. I want to load two kernel modules ("**ashmem**" and "**binder**") with Secure Boot enabled. At least for me, "binder" loads fine with Secure Boot enabled, but "ashmem" does ***not*** load with Secure Boot enabled. *Both* modules load with Secure Boot disabled.

To the above, I’ll add that I had enabled Secure Boot (and enrolled MOK) during the *initial* clean  installation of Ubuntu 20.04. Even so, following the *excellent* suggestion provided by [Rinzwind](https://askubuntu.com/users/15811/rinzwind), I ran the following *after* the initial installation:

```
sudo kmodsign sha512 /var/lib/shim-signed/mok/MOK.priv /var/lib/shim-signed/mok/MOK.der /lib/modules/`uname -r`/kernel/drivers/staging/android/ashmem_linux.ko
```
And after running the above, I re-ran:

```
sudo modprobe ashmem_linux

sudo modprobe binder_linux

lsmod | grep -e ashmem_linux -e binder_linux
```

SUCCESS!  Both “ashmem” and “binder” loaded!  Thanks [Rinzwind](https://askubuntu.com/users/15811/rinzwind)!!

What’s more, [Anbox](https://anbox.io/) starts for me now with Secure Boot enabled, which was my end goal.  :)

# Request #

@morphis 

I hope that you will please consider updating the Anbox installer to (when installing Anbox onto Ubuntu 20.04 with Secure Boot enabled) *automatically* run the following or similar:

```
sudo kmodsign sha512 /var/lib/shim-signed/mok/MOK.priv /var/lib/shim-signed/mok/MOK.der /lib/modules/`uname -r`/kernel/drivers/staging/android/ashmem_linux.ko
```
Also, please note that whether Secure Boot is enabled can be determined using the `sudo mokutil --sb-state` command.
For example, running `sudo mokutil --sb-state` on a computer having Secure Boot enabled should yield:
```
$ sudo mokutil --sb-state
[...]

PR #1588, opened 2020-08-28
Update instructions for Ubuntu 20.04

JonasVautherin:
Update documentation following https://github.com/anbox/anbox/issues/1579 and https://github.com/anbox/anbox/issues/1570. Thanks @GizmoChicken for the help!

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -33,9 +33,9 @@ everything you need to run the full Anbox experience.
 
 ## Install necessary kernel modules
 
-### On Ubuntu 20.04
+### On Ubuntu 19.04 and later
 
-There is no need to install those modules on Ubuntu 20.04. However, you may get an error loading `ashmem_linux` if SecureBoot is enabled:
+There is no need to install those modules on a recent Ubuntu or Debian: Even Ubuntu 18.04 LTS and Debian Stable have got the modules. However, you may get an error loading `ashmem_linux` if SecureBoot is enabled:
 
 ```
  $ sudo modprobe ashmem_linux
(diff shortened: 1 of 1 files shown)


necessarily-equal:
Hi. Thanks for your contribution.

It seems that the out of tree kernel modules stopped working in 19.04 instead of 20.04. The modules were in the Ubuntu (and Debian) kernel for longer. I'd personnally put the recent information first, then say that if they've got an old kernel, they can try the other way.

docs/install.md
```diff
@@ -33,6 +33,27 @@ everything you need to run the full Anbox experience.
 
 ## Install necessary kernel modules
 
+### On Ubuntu 20.04
+
+There is no need to install those modules on Ubuntu 20.04. However, you may get an error loading `ashmem_linux` if SecureBoot is enabled:
```
necessarily-equal, line 38: I'd do this
```suggestion
There is no need to install those modules on a recent Ubuntu or Debian: Even Ubuntu 18.04 LTS and Debian Stable have got the modules. However, you may get an error loading `ashmem_linux` if SecureBoot is enabled:
```

docs/install.md
```diff
@@ -33,6 +33,27 @@ everything you need to run the full Anbox experience.
...
+
+```
+$ sudo mokutil --sb-state
+SecureBoot enabled
+```
+
+There are two ways around this. One is to disable the SecureBoot: https://wiki.ubuntu.com/UEFI/SecureBoot/DKMS. 
+Following [this post](https://github.com/anbox/anbox/issues/1570), the other way is to sign the `ashmem_linux` kernel module yourself. Note that you may have to enroll your own key, as described [here](https://ubuntu.com/blog/how-to-sign-things-for-secure-boot).
+
+### Before Ubuntu 20.04
```
necessarily-equal, line 55:
```suggestion
### Before Ubuntu 19.04
```

docs/install.md
```diff
@@ -33,6 +33,27 @@ everything you need to run the full Anbox experience.
...
+
```
necessarily-equal, line 56:
```suggestion

You should have the modules on your system already. However, you can still try to use the ppa (no longer maintained) if `sudo modprobe ashmem_linux binder_linux` can't find the modules.

```

PR #1594, opened 2020-09-02
More sensors support

tomasz-grobelny:
Add support for more Goldfish sensors
Add documentation about DBUS interface
Make Anbox compile with Clang 9.0 on ARM


tomasz-grobelny:
More sensors to come, but comments already welcome (@necessarily-equal, @morphis).

necessarily-equal (changes requested):

src/anbox/qemu/sensors_message_processor.cpp
```diff
@@ -63,7 +75,6 @@ void SensorsMessageProcessor::handle_command(const string &command) {
 void SensorsMessageProcessor::send_message(const string &msg) {
   send_header(msg.length());
   messenger_->send(msg.c_str(), msg.length());
-  finish_message();
```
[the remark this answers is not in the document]
tomasz-grobelny: The format of the message is length+data+length+data+.... The send_header method sends length (string length) and then we send data. If finish_message sends null it is then interpreted as part of length of next packet. The string null termination is done on receiving end so no need to send it here.

src/anbox/cmds/check_features.cpp
```diff
@@ -57,6 +57,7 @@ anbox::cmds::CheckFeatures::CheckFeatures()
           cli::Description{"Check that the host system supports all necessary features"}} {
 
   action([this](const cli::Command::Context&) {
+    (void)this;
```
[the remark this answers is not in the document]
tomasz-grobelny: Clang complains about unused lambda capture.

CMakeLists.txt
```diff
@@ -62,6 +62,7 @@ find_package(Threads)
 find_package(EGL REQUIRED)
 find_package(GLESv2 REQUIRED)
 find_package(Protobuf REQUIRED)
+find_package(fmt REQUIRED)
```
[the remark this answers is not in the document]
tomasz-grobelny: Done.

tomasz-grobelny:
This pull request is complete. Please review (@morphis, @necessarily-equal).

necessarily-equal (approved):

docs/dbus.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Launching applications
+Launching applications is done using `org.anbox.ApplicationManager` DBUS interface. To launch application you can use the following command:
+```
+anbox launch --package=org.anbox.appmgr --component=org.anbox.appmgr.AppViewActivity
+```
+This command internally uses DBUS to contact session manager and calls `Launch` method to launch specified component. Unfortunately due to limitations of `dbus-send` (in particular limited argument type support) this cannot be done using the generic dbus command line interface.
+
+## Sensor values
+You can get or set sensor values using `org.anbox.Sensors` interface.
```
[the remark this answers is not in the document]
tomasz-grobelny: Added list of available sensors and Python example.

tomasz-grobelny:
@morphis, any thoughts?

src/anbox/qemu/sensors_message_processor.cpp
```diff
@@ -17,24 +17,64 @@
...
     shared_ptr<network::SocketMessenger> messenger, shared_ptr<application::SensorsState> sensorsState)
     : QemudMessageProcessor(messenger), sensors_state_(sensorsState) {
+  enabledSensors_ = 0;
   thread_ = std::thread([this]() {
     for (;;) {
-      if (temperature_.load())
-        send_message("temperature:" + to_string(sensors_state_->temperature));
+      auto enabledSensors = enabledSensors_.load();
+      if (enabledSensors & SensorType::AccelerationSensor)
+        send_message(fmt::format("acceleration:{0}", sensors_state_->acceleration));
```
morphis, line 58: Any reason we're using fmt::format when we already have `string_format` in `src/anbox/utils.h`? Using different implementations for the same thing isn't what we should do.

morphis (changes requested):

src/anbox/qemu/sensors_message_processor.cpp
tomasz-grobelny: Hmm... Reason was that I didn't know about `string_format`. I can refactor if required. On the other hand seems like libfmt is the future of string formatting in C++ that made it into recent standard (C++20). Decision is yours.

tomasz-grobelny: @morphis, how do we proceed with this PR? Should I make the changes to use `string_format` or can it be merged as is?

morphis: I would like our code to be consistent. So either we convert evertyhing to `std::format` or we stick to `string_format` for the time being. Given the amount of work a conversion is, I would like to stick to `string_format` in this PR and then we can move globally to something else.

tomasz-grobelny: Done.

morphis:
Thanks! LGTM

PR #1597, opened 2020-09-06
Use current system user and group in docker build script

jnnksdev:
When compiling in a container the system user root is owning all of the produced artifacts.
Moving and removing build artifacts was not possible without elevation.

The script was modified to use the current system user and group id instead.


morphis:
LGTM, thanks!

PR #1599, opened 2020-09-08
snap: drop lxc fork

necessarily-equal:
The LXC fork that is used by the snap package hasn't been updated for some years and is currently not being used anyway (Snap will try to use the host system's liblxc.so.1!) This PR attemps to make Anbox work with Ubuntu's LXC package for the snap.

However, it doesn't work yet as it fails with this error in `container.log`:

```
lxc 20200908144116.511 INFO     start - start.c:do_start:1211 - Unshared CLONE_NEWCGROUP
lxc 20200908144116.511 TRACE    conf - conf.c:remount_all_slave:3094 - Remounted all mount table entries as MS_SLAVE
lxc 20200908144116.511 ERROR    conf - conf.c:lxc_mount_rootfs:1242 - No such file or directory - Failed to access to "/usr/lib/x86_64-linux-gnu/lxc". Check it is present
lxc 20200908144116.511 ERROR    conf - conf.c:lxc_setup_rootfs_prepare_root:3178 - Failed to setup rootfs for
lxc 20200908144116.511 ERROR    conf - conf.c:lxc_setup:3277 - Failed to setup rootfs
lxc 20200908144116.511 ERROR    start - start.c:do_start:1231 - Failed to setup container "default"
lxc 20200908144116.511 ERROR    sync - sync.c:__sync_wait:40 - An error occurred in another process (expected sequence number 5)
lxc 20200908144116.511 DEBUG    network - network.c:lxc_delete_network:3693 - Deleted network devices
lxc 20200908144116.511 TRACE    start - start.c:lxc_serve_state_socket_pair:491 - Sent container state "ABORTING" to 13
lxc 20200908144116.511 TRACE    start - start.c:lxc_serve_state_clients:427 - Set container state to ABORTING
```

Any idea?

### DIFF
--- a/scripts/container-manager.sh
+++ b/scripts/container-manager.sh
@@ -45,8 +45,8 @@ start() {
 	# Ensure FUSE support for user namespaces is enabled
 	echo Y | tee /sys/module/fuse/parameters/userns_mounts || echo "WARNING: kernel doesn't support fuse in user namespaces"
 
-	# liblxc.so.1 is in $SNAP/lib
-	export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$SNAP/liblxc"
+	# liblxc.so.1 is in $SNAP/usr/lib/$ARCH-linux-gnu
+	export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$SNAP/usr/lib/$ARCH-linux-gnu"
 
 	# For unknown reason we got bug reports that the container manager failed to start
 	# because it cannot find libboost_log.so.1.58.0 To mitigate this we're adding the
--- a/scripts/snap-wrapper.sh
+++ b/scripts/snap-wrapper.sh
@@ -13,10 +13,10 @@ fi
 # where it can be found
 export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$SNAP/usr/lib/$ARCH/pulseaudio"
 
-# liblxc.so.1 is in $SNAP/usr/lib
-export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$SNAP/lib:$SNAP/usr/lib"
+export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$SNAP/lib"
 
-export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$SNAP/usr/lib/$ARCH"
+# liblxc.so.1 is in /usr/lib/$ARCH-linux-gnu/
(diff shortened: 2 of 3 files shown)


GizmoChicken:
>Any idea?

No ideas from me.   But I thank you for looking into this!  Much appreciated!!

jnnksdev:
`Failed to access to "/usr/lib/x86_64-linux-gnu/lxc". Check it is present`

It is created when installing LXC (as in `apt install lxc`). There is a README inside the directory:
```
This directory must exist, even though it may be empty.
It is used to temporary mount the rootfs of lxc in a private
mount namespace only visible by the processes running in the
container.
```
That would explain `Failed to setup rootfs...`.

Shouldn't anbox look in `/snap/anbox/69/lib/x86_64-linux-gnu/lxc` or so instead? Seems like it still wants to use system libs instead of the one packaged with snap.

necessarily-equal:
Ah, yes. I've seen that on Alpine but with a different directory. I'm wondering if this must be set at compilation time or if it is possible to set that with an environment variable.

If not, I guess the easiest is to create the directory if it doesn't exist already. That's ugly, but it's still much better than the current situation where the snap gets the lxc code from the outside system…

jnnksdev:
Well the README did not tell the entire story, as the directory also contains some binaries:
```
user@bionic:~$ ls -op /usr/lib/x86_64-linux-gnu/lxc/
total 212
drwxr-xr-x 2 root   4096 Sep 11 18:13 hooks/
-rwxr-xr-x 1 root    521 Nov 23  2018 lxc-apparmor-load
-rwxr-xr-x 1 root   3090 Nov 23  2018 lxc-containers
-rwxr-xr-x 1 root  84360 Nov 23  2018 lxc-monitord
-rwxr-xr-x 1 root   9417 Nov 23  2018 lxc-net
-rwsr-xr-x 1 root 100760 Nov 23  2018 lxc-user-nic
-rw-r--r-- 1 root    189 Nov 23  2018 README
```
Just creating the directory won't do much as these global services are missing.
_____

In `conf.c` you can find the relevant line 1242/1244 checking for the directories existence. Tracing back `rootfs->mount` I found only one spot where it is set to an actual value (`conf.c: 2585`). Going further back, AM_CFLAG `LXCROOTFSMOUNT`(`conf.c: 2554`) is used for that and controlled by the environment variable with the same name (`Makefile.am:202`, created when [initiating a build](https://github.com/lxc/lxc#building-lxc)). 
I can't trace the setting of the env variable, but when compiling, it's value is `LXCROOTFSMOUNT="/usr/local/lib/lxc/rootfs"`, which is translated by the linker to `/usr/lib/x86_64-linux-gnu/lxc`.

**TLDR**: value is set at compilation time

PR #1606, opened 2020-09-13
Allow injecting GPS location via DBus calls

tomasz-grobelny:
Tested to work on data generated from https://nmeagen.org/ and OsmAnd. Ready for review (@morphis, @necessarily-equal).

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ $ sudo apt install build-essential cmake cmake-data debhelper dbus google-mock \
     libproperties-cpp-dev libprotobuf-dev libsdl2-dev libsdl2-image-dev lxc-dev \
     pkg-config protobuf-compiler python-minimal
 ```
-We recommend Ubuntu 18.04 (bionic) with **GCC 7.x** as your build environment.
+We recommend Ubuntu 20.04 (focal) with **GCC 7.x** as your build environment.
 
 
 ### Build
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ $ sudo apt install build-essential cmake cmake-data debhelper dbus google-mock \
     libproperties-cpp-dev libprotobuf-dev libsdl2-dev libsdl2-image-dev lxc-dev \
     pkg-config protobuf-compiler python-minimal
 ```
-We recommend Ubuntu 20.04 (focal) with **GCC 7.x** as your build environment.
+We recommend Ubuntu 20.04 (focal) as your build environment.
 
 
 ### Build
--- a/src/anbox/qemu/pipe_connection_creator.h
+++ b/src/anbox/qemu/pipe_connection_creator.h
@@ -23,6 +23,7 @@
 #include <memory>
 
 #include "anbox/application/sensors_state.h"
+#include "anbox/application/gps_info_broker.h"
 #include "anbox/do_not_copy_or_move.h"
 #include "anbox/network/connection_creator.h"
 #include "anbox/network/connections.h"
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -108,7 +108,7 @@ $ sudo apt install build-essential cmake cmake-data debhelper dbus google-mock \
     libproperties-cpp-dev libprotobuf-dev libsdl2-dev libsdl2-image-dev lxc-dev \
     pkg-config protobuf-compiler python-minimal
 ```
-We recommend Ubuntu 18.04 (bionic) with **GCC 7.x** as your build environment.
+We recommend Ubuntu 20.04 (focal) with **GCC 7.x** as your build environment.
```
morphis, line 111: Do we still recommend gcc 7.x on 20.04?

src/anbox/qemu/pipe_connection_creator.h
```diff
@@ -55,6 +56,7 @@ class PipeConnectionCreator
     qemud_gsm,
     qemud_adb,
     bootanimation,
+    qemud_gps,
```
morphis, line 59: Can we put it in order with the other `qemud_*` ones?

morphis (changes requested):

src/anbox/qemu/pipe_connection_creator.h
tomasz-grobelny: Sure, can be done.

README.md
tomasz-grobelny: I am using Clang 9.0 (to use clangd). But gcc 7.5.0 is also fine. Gcc 9.3.0 should also be fine. Should I remove compiler recommendation?

tomasz-grobelny: Removed info about gcc 7.x.

tomasz-grobelny:
@morphis, is there anything more that needs to be fixed before merge?

zvova7890:
Hey, cool feature, thank you! I've found some small deinitialization issue, when shutdown session it segfault if we send coordinates repeatedly via dbus. As I'm understand, something has been deleted(like a messenger_) but still in use   gps_info_broker_ signals.

`
0  0x00007ffff7540615 in raise () from /usr/lib/libc.so.6
#1  0x00007ffff7529862 in abort () from /usr/lib/libc.so.6
#2  0x00007ffff777c86a in __gnu_cxx::__verbose_terminate_handler () at /build/gcc/src/gcc/libstdc++-v3/libsupc++/vterminate.cc:95
#3  0x00007ffff7788d9a in __cxxabiv1::__terminate (handler=<optimized out>) at /build/gcc/src/gcc/libstdc++-v3/libsupc++/eh_terminate.cc:48
#4  0x00007ffff7788e07 in std::terminate () at /build/gcc/src/gcc/libstdc++-v3/libsupc++/eh_terminate.cc:58
#5  0x00007ffff7789b65 in __cxxabiv1::__cxa_pure_virtual () at /build/gcc/src/gcc/libstdc++-v3/libsupc++/pure.cc:50
#6  0x0000555555a3bce1 in operator() (__closure=0x7fffac000c10, sentence="$GPGGA,160804,4023.1171,N,1028.8062,E,1,12,0,0.0,M,0.0,M,,*50")
    at /tmp/anbox-git/src/anbox/src/anbox/qemu/gps_message_processor.cpp:29
`

mikhailnov:
Can you please provide and example `dbus-send` command to set location via this dbus interface?

necessarily-equal:
https://github.com/anbox/anbox/blob/master/docs/dbus.md

mikhailnov:
> https://github.com/anbox/anbox/blob/master/docs/dbus.md

There is no example for Gps there, the interface is a bit different and the format is not documented...

tomasz-grobelny:
> > https://github.com/anbox/anbox/blob/master/docs/dbus.md
> 
> There is no example for Gps there, the interface is a bit different and the format is not documented...

    while IFS= read -r line
    do
        echo "$line"
        dbus-send --session --dest=org.anbox --type=method_call --print-reply /org/anbox org.anbox.Gps.PushSentence "string:$line"
    done < <(gpspipe -r|stdbuf -i0 -o0 -e0 grep "\$GP")

This one works for me - it pushes full NMEA sentences from system gpsd to anbox.

mikhailnov:
Thanks, could you please show an example output of gpspipe -r ?

tomasz-grobelny:
> Thanks, could you please show an example output of gpspipe -r ?

It would look like this:

    $GPGGA,123519,4807.038,N,01131.000,E,1,08,0.9,545.4,M,46.9,M,,*47

(from Wikipedia). https://nmeagen.org/ might be useful for testing.

PR #1618, opened 2020-09-23
do not automatically hard reset git tree on build

jnnksdev:
Added optional parameter to `build.sh` and `build-with-docker.sh` to prevent auto git clean and reset


jnnksdev:
I overlooked `.travis.yml` and `.github/workflows/main.yml`, which are using `build-with-docker.sh` without `--clean`.

Can I still modify this PR somehow, or do I need to open another?

PR #1629, opened 2020-10-06
platform/sdl: Disable mouse event emulation for touch input

dos1:
SDL 2.0.10 introduced mouse event emulation for touch input.
However, Anbox already handles touch events by itself, so this just
results in duplicated events and erroneous input when using Anbox
with a touch screen (for instance, on the Librem 5).

Signed-off-by: Sebastian Krzyszkowiak <sebastian.krzyszkowiak@puri.sm>


morphis:
LGTM, thanks!

Issue #1628, opened 2020-10-05
Building enhancement


uis246:
1. Remove libsystemd from dependencies. Most of Gentoo and Artix users use OpenRC.
2. Make tests optional

Also add properties-cpp to requirements list

okias:
1. highly unlikely it'll happen [1] (also read the commit message), but possibility could be make it optional. Now it's up to interested people (OpenRC, Alpine/pmOS guys) to present solution or patches.
2. that would make sense, at least for distributions as Gentoo. Also I assume that's not so hard to implement.

[1] https://github.com/anbox/anbox/commit/f1480bfb914bee4a0a564b9e67deab51342684f6

uis246:
Why sd_bus instead freedesktop's one? Just libdbus. Not dbus-cpp.

necessarily-equal:
openRC users can still use Anbox if they install elogind instead. Anbox recently moved (https://github.com/anbox/anbox/pull/1540) from sdbus to sdbus-cpp and that allowed dropping quite some code. As the Alpine maintainer for the Anbox package, I didn't oppose it, though it could have some cleanups to avoid bundling sdbus-cpp if the distro packages it already.

ShadowJonathan:
https://github.com/anbox/anbox/issues/1709

ShadowJonathan:
https://github.com/anbox/anbox/issues/1709

PR #1632, opened 2020-10-06
Make tests optional

uis246:
Second from #1628


uis246:
For testing enable "CMAKE_TESTING_ENABLED" flag

CMakeLists.txt
```diff
@@ -138,7 +135,9 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/anbox/build/config.h.in
 
 add_subdirectory(external)
 add_subdirectory(src)
-add_subdirectory(tests)
+if (CMAKE_TESTING_ENABLED)
```
morphis, line 138: This disables the tests by default which breaks our CI. By default tests should be enable and if somebody wants them disabled (e.g for cross compilation) the flag can be set.

morphis (changes requested):

CMakeLists.txt
uis246: Most opensource projects keep tests disabled by default.

necessarily-equal:
Note that we do have BUILD_TESTING because cpu_features define it. Maybe we should reuse the name?

uis246:
> Note that we do have BUILD_TESTING because cpu_features define it. Maybe we should reuse the name?

Thanks

uis246:
Done

CMakeLists.txt
morphis: Why would tests be disabled by default? Whenever you build something you always want it to be tested. Otherwise you cannot trust what you've build.

uis246: I replaced CMAKE_TESTING_ENABLED to BUILD_TESTING defined in cpu_features.

necessarily-equal:
still useful

uis246:
!

PR #1647, opened 2020-10-22
use python3

jnnksdev:
* upgraded gen-emugl-entries from [googlesource](https://android.googlesource.com/platform/external/qemu/+/emu-master-dev/android/scripts/gen-entries.py "gen-entries.py") to use python3
* added dependencies to docker, snapcraft and readme

The generated headers are equal.


morphis:
LGTM, thanks!

PR #1660, opened 2020-10-31
session-manager --window-size is ignored

jnnksdev:
The passed value is ignored and is always the default.

Presumably an issue with argument instantiation. I suppose the current cli flag implementation cannot handle complex objects like `graphics::Rect`.

* convert argument `window-size` to string
* parse `window-size=[width],[height]` argument when called with `--single-window`
  * give format hints when exceptions occur


jnnksdev:
nevermind...

PR #1665, opened 2020-11-03
include googletest as submodule

jnnksdev:
CMake had problems finding the gtest/gmock libs every now and then.
The problem seems to be non-deterministic with random occurrences.

Include the source repo as submodule, instead of depending on an installed library.

*PR might break when https://github.com/anbox/anbox/pull/1632 is merged.*


CameronNemo:
I would greatly prefer using more system libraries, not less. Embedded libraries make compilation take longer and can allow bugs to lurk.

jnnksdev:
We share the same perspective in terms of dependencies. I too would like to outsource as much as possible and create a stable build environment instead. Unfortunately a lot of people had problems with breaking builds, because googletest/gtest/gmock was not found. I had my share of problems as well.
Before investigating why it is not reliable as a system lib and think about improvements, we need to make sure it works in the first place. A slower build is better than no build.

In this case the compilation time is increased by 6.5s (on an old 4c/8t CPU). I can accept this increase especially in relation to the overall build time of 2m20s. And even then compiling only takes longer for the first time, after that it does not make a difference anymore, as long as you rebuild.

CameronNemo:
Could you at least make it an option to prefer the system version of google test and other submodules?

jnnksdev:
Ok, then I will do more research about CMakes dependency management then.
How about this:
* local googletest v1.10 ?
  * yes: use system library
  * no: clone & compile

CameronNemo:
I don't think a warning is 100% necessary, but yeah that sounds nice.

morphis:
I second @CameronNemo here. Where we can, we should use system libraries rather than adding more submodules. Ideally we use zero submodules and take everything from the system. For gtest the situation is a bit awkward as @jnnksdev described and we indeeded at multiple people having problems with this on their system.

So using the approach to prefer the system gtest and if not available fallback to the submodule is good.

necessarily-equal:
note that hopefully many of these issues were fixed by https://github.com/anbox/anbox/pull/1423. I personally have only saw people that didn't install gtest have issue with it since. Installing the package fixes it, and if it doesn't, then people can always use `-DBUILD_TESTING=DISABLE` (that needs some knowledge tough, but so does installing from source in the first place)

jnnksdev:
After thinking about this for some time I would like to close this PR and leave the behavior as is.
Instead the docker build container should be promoted as the recommended build solution.

PR #1668, opened 2020-11-05
Refactor namespaces declarations

jnnksdev:
C++17 introduced nested namespaces to reduce declaration footprint.
From:
```
namespace first {
    namespace second { 
        ... 
    }
}
```
To:
```
namespace first::second {
    ...
}
```
A simple syntax change that improves readability.
This refactoring is related to upcoming experiments regarding [C++20 modules](https://en.cppreference.com/w/cpp/language/modules).

Since the semantics have not been changed, I would argue that copyright headers in code files are still valid.

Somehow the googletest submodule was able to sneak into this branch. This PR is still up for discussion anyway, so I would wait until submodule PR is merged and then figure out an appropriate fix for that.

### DIFF
--- a/src/anbox/android/intent.cpp
+++ b/src/anbox/android/intent.cpp
@@ -19,8 +19,7 @@
 
 #include <ostream>
 
-namespace anbox {
-namespace android {
+namespace anbox::android {
 bool Intent::valid() const {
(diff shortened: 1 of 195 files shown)


jnnksdev:
will close, posted another PR without submodule issue

PR #1670, opened 2020-11-06
delete unused file fd_sets.h

jnnksdev:
File only contained `typedef std::vector<std::vector<Fd>> FdSets;` which is unused.


jnnksdev:
will create a more detailed PR with all unused files

PR #1671, opened 2020-11-06
delete unused files mir_display_connection.*

jnnksdev:
Files were not included anywhere (not even in CMakeFiles)


jnnksdev:
will create a more detailed PR with all unused functions

PR #1672, opened 2020-11-06
remove unused function write_file_at and write_to_fd

jnnksdev:
write_to_fd was only called once in write_file_at, which was unused.


jnnksdev:
will create a more detailed PR with all unused functions

PR #1677, opened 2020-11-10
remove PendingCallCache from ManagementApiSkeleton

jnnksdev:
PendingCallCache was a ctor param of ManagementApiSkeleton, but was not used apart from setting the private member.


morphis:
LGTM, thanks!

PR #1682, opened 2020-11-17
remove unused parameter Runtime

jnnksdev:
`rpc::ConnectionCreator` has unused reference to `Runtime` as parameter

* remove parameter


morphis:
LGTM, thanks!

PR #1700, opened 2020-12-09
Do break out of while loop

lilose:
Better to do break out of while loop.


lilose:
@morphis would you please help to teach me how to add you to reviewers? thanks~~~

morphis:
LGTM, thanks!

PR #1733, opened 2021-01-21
Optionally use installed sdbus-cpp library

jnnksdev:
This PR implements a compile-once solution for the external project `sdbus-cpp` used for the docker container build toolchain.
The installed library is used instead of building the in-tree module when it can be found with cmake `find_package`.
A fallback was implemented to avoid build errors in case the library is not installed on the host.

This is a small study to reduce build times by providing precompiled resources before Anbox is built.
The performance improvement is for niche use cases where the entire project is fully rebuild multiple times.

By introducing other concepts, the overall build time is reduced anyways. For example docker *BuildKit*, which reduces the buld time for the `anbox-build` image.

|job|master (s)|this PR (s)|diff (master - PR in s)|
|---|---|---|---|
|scripts/build.sh|60.41|59.51|0.9|
|scripts/build-with-docker.sh|61.28|60.15|1.13|
|scripts/build-with-docker.sh (docker --no-cache)|163.44|156.99|6.45|

When accepted this will be a reference implementation with the other external modules following in the same style.

@morphis In  `src/CMakeLists.txt` the include path for `sdbus-cpp` is hardcoded. I would like to use something like `sdbus-c++_INLCUDE_DIRS`, but the variable was empty. Do you know a work-around?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,6 +136,22 @@ set(ANBOX_STATEDIR_FULL "${CMAKE_INSTALL_FULL_LOCALSTATEDIR}/lib/anbox")
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/anbox/build/config.h.in
                ${CMAKE_CURRENT_SOURCE_DIR}/src/anbox/build/config.h)
 
+
+# sdbus-cpp
+#   find includes and code-generator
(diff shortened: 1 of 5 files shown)


Fuseteam:
What is this missing?

Fuseteam:
Still sounds useful...

PR #1734, opened 2021-01-22
New Build Script

jnnksdev:
Merges the two build scripts `scripts/build.sh` and `scripts/build-with-docker.sh` into one and moves it to root directory.

New CLI:
```
build.sh [options] -- [cmake args]

Options:
--clean      : remove contents of directory ./build
               fully rebuild the project
--hard-reset : reset current directory to last commit, implies --clean
               In cases where anbox comes directly from a checked out Android
               build environment we miss some symlinks which are present on the
               host and don't have a valid git repository in that case.
--docker     : use a docker container to build Anbox
-j, --jobs=X : number of parallel make jobs, default is 8
               equal to 'make --jobs=X'
```

@morphis Is `--hard-reset` still relevant? I would like to remove it, as it is not directly connected to the build task.
Maybe there is a better way to solve the problem than this.


morphis:
LGTM, thanks!

PR #1750, opened 2021-02-14
Do not install cpu_features on make install

thotypous:
We found that, when packaging anbox for Arch Linux (https://aur.archlinux.org/packages/anbox-git/), it conflicts with another package (libvolk). This is because both packages ship with cpu_features.

It is completely unnecessary to have cpu_features installed for Anbox to work, since it is a static library and is already linked into Anbox anyway.

Therefore, this pull requests disables installation of cpu_features on `make install`.


morphis:
LGTM, thanks!

PR #1752, opened 2021-02-22
Update fd_socket_transmission.cpp

nghuyy:


src/anbox/network/fd_socket_transmission.cpp
```diff
@@ -69,7 +69,9 @@ void send_fds(Fd const& socket, std::vector<Fd> const& fds) {
     message->cmsg_level = SOL_SOCKET;
     message->cmsg_type = SCM_RIGHTS;
 
-    int* const data = reinterpret_cast<int*>(CMSG_DATA(message));
+    unsigned char* msgtp = CMSG_DATA(message);
+    int* const data = reinterpret_cast<int*>(&msgtp);
```
morphis, line 73: What is the purpose of this change?

morphis (changes requested):

src/anbox/network/fd_socket_transmission.cpp
IsioNisin: This changes are necessary for compilation on some armhf devices. In my case it was Exynos 4412 SoC device, and i have a cast-align error in this line and also line 147.

PR #1763, opened 2021-03-08
Improve performance

mariogrip:
This removes unnecessary std::copy

For the amount these functions get called, std copy is way to expensive
to do. This removes the need for buffer to be copied twice, first to a
std::vector then SmallFixedVector. This also directly copies buffer into
a smallVector without the init overhead smallVector's ctor does. This
result is a massive performance boost (at least on arm devices).

Also use memcpy interestingly seems to be faster on arm
devices, on amd64 they seems to be about the same speed.

Same as https://github.com/anbox/anbox/pull/902 but 2 years late :)


morphis (approved):
LGTM

PR #1765, opened 2021-03-09
Add "rootless window mode"

mariogrip:
This adds rootless mode, in this mode a window wont be created until a
application is started.


morphis (approved):
LGTM, thanks!

Issue #1707, opened 2020-12-17
Failed to build Anbox: gmock


Chaosdave33:
I've get an error while building anbox:
`pi@raspberrypi:~/Workspace/anbox/build $ make
[ 90%] Built target GMock
[ 90%] Built target sdbus-cpp
[ 90%] Built target process-cpp
[ 90%] Built target emugen
[ 90%] Built target OpenglCodecCommon
[ 90%] Built target GLESv2_dec
[ 90%] Built target GLESv1_dec
[ 90%] Built target emugl_common
[ 90%] Built target OpenGLESDispatch
[ 90%] Built target renderControl_dec
[ 90%] Built target xdg
[ 90%] Built target xdg_test
[ 90%] Built target backward_object
[ 90%] Built target backward
[ 90%] Built target unix_based_hardware_detection
[ 90%] Built target utils
[ 90%] Built target cpu_features
[ 90%] Built target list_cpu_features
[ 90%] Built target anbox-protobuf
[ 90%] Built target anbox-core
[ 90%] Built target anbox
make[2]: *** Keine Regel vorhanden, um das Ziel „gmock/lib/libgmock.a“, 
  benötigt von „tests/anbox/android/intent_tests“, zu erstellen.  Schluss.
make[1]: *** [CMakeFiles/Makefile2:2216: tests/anbox/android/CMakeFiles/intent_tests.dir/all] Fehler 2
make: *** [Makefile:141: all] Fehler 2
`
Building on aarch64 with libgmock-dev installed.

linux0ne-su:
make[2]: *** 没有规则可制作目标“gmock/lib/libgmock.a”，由“tests/anbox/android/intent_tests” 需求。 停止。
CMakeFiles/Makefile2:2045: recipe for target 'tests/anbox/android/CMakeFiles/intent_tests.dir/all' failed
make[1]: *** [tests/anbox/android/CMakeFiles/intent_tests.dir/all] Error 2
Makefile:159: recipe for target 'all' failed
make: *** [all] Error 2

我也遇到这样的问题，请问这该怎么解决 ？

Issue #1662, opened 2020-10-31
Unable to build from source (No rule to make target 'gmock/lib/libgmock.a')


bidaian:
I followed the **Build from source** instructions in README.md:

```
$ sudo apt install build-essential cmake cmake-data debhelper dbus google-mock     libboost-dev libboost-filesystem-dev libboost-log-dev libboost-iostreams-dev     libboost-program-options-dev libboost-system-dev libboost-test-dev     libboost-thread-dev libcap-dev libexpat1-dev libsystemd-dev libegl1-mesa-dev     libgles2-mesa-dev libglm-dev libgtest-dev liblxc1     libproperties-cpp-dev libprotobuf-dev libsdl2-dev libsdl2-image-dev lxc-dev     pkg-config protobuf-compiler python3-minimal
$ git clone https://github.com/anbox/anbox.git --recurse-submodules
$ cd anbox
$ mkdir build
$ cd build
$ cmake ..
$ make

```
But at the make stage this is the error I get:
```

[ 40%] Built target anbox-protobuf
[ 89%] Built target anbox-core
[ 90%] Built target anbox
make[2]: *** No rule to make target 'gmock/lib/libgmock.a', needed by 'tests/anbox/android/intent_tests'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:2215: tests/anbox/android/CMakeFiles/intent_tests.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

vrichomme:
Look to my bug report here, I had the same issue: https://github.com/anbox/anbox/issues/1656

jnnksdev:
just sent a PR about this: https://github.com/anbox/anbox/pull/1665

FCLC:
can be fixed by adding cmake-extras to your installed libraries. 

I'm making a pull request RN to mitigate this issue. 

In more detail, because cmake-helper isnt explicitly "needed" you don't get notified that the issue is finding the library per se, since you "know" that the library itself is installed.

Fuseteam:
this has an open pr

Issue #1616, opened 2020-09-22
Help.


Innomenatus:
make[2]: *** No rule to make target 'gmock/lib/libgmock.a', needed by 'tests/anbox/android/intent_tests'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:2215: tests/anbox/android/CMakeFiles/intent_tests.dir/all] Error 2
make: *** [Makefile:141: all] Error 2

6. ** Please paste the result of `anbox system-info` below:**
<details>
<summary>anbox system-info output</summary>

```
[please paste printout of `anbox system-info` here]
```
</details>

**Please describe your problem:**

**What were you expecting?:**

**Additional info:**

jnnksdev:
On some distros you need to install `libgmock-dev`.
I would highly recommend to use the provided Docker container to build Anbox. You can start a build with `scripts/build-with-docker.sh`. Be aware that the script will hard reset your git tree (for now, change is on the way).

Innomenatus:
I installed libgmock-dev already. How do you start a build with scripts/build-with-docker.sh? I'm new here.

jnnksdev:
When you clone the repository there is a folder called `scripts`. In this folder you will find a shell script called `build-with-docker.sh` that builds and starts a docker container. Anbox will be compiled inside the container where all dependencies are present without having to install anything.

You need to be in the root folder of the repository and execute `scripts/build-with-docker.sh` in a shell window.

May I ask why you want to compile the code?

necessarily-equal:
This might help https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/testing/anbox/fix-gtest-use.patch though it doesn't seem to touch the same part of the code.

PR #1258, opened 2019-10-11
use dockerfile

stephengroat:
starts moving towards using a `Dockerfile` over other types of docker builds

I registered the docker org `anbox` (more than willing to transfer it over), but you can easily take 1-2 mins off of the clean builds by uploading the image to the Hub.

PR #1778, opened 2021-03-27
Added cmake-extras to dependencies- Fixes some build from source issues

FCLC:
deals with cmake not finding installed libraries such as gmock 

#1707 #1662 #1616 

TLDR; it seems that the way things were set up made the assumption that cmakes-extras was installed. 

this issue was found when comparing docker dependencies in  #1723 #1258

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ $ sudo apt install build-essential cmake cmake-data debhelper dbus google-mock \
     libboost-thread-dev libcap-dev libexpat1-dev libsystemd-dev libegl1-mesa-dev \
     libgles2-mesa-dev libglm-dev libgtest-dev liblxc1 \
     libproperties-cpp-dev libprotobuf-dev libsdl2-dev libsdl2-image-dev lxc-dev \
-    pkg-config protobuf-compiler python3-minimal
+    pkg-config protobuf-compiler python3-minimal cmake-extras
 ```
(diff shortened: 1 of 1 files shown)


FCLC:
The above update to the pull request adds missing dependencies for certain distribution of ubuntu, such as certain arm and aarch64 LTS builds

FCLC:
Added troubleshooting steps to building android to deal with the typical default of python 2 on some distributions of ubuntu

README.md
```diff
@@ -122,7 +122,23 @@ $ cd anbox
 $ mkdir build
 $ cd build
 $ cmake ..
-$ make
+$ make -j$(nproc)
+```
+#### cmake error: no cpu_features
```
morphis, line 127: Can we move this to `docs/troubleshooting.md` and link here?

morphis (changes requested):

Theslees:
this is extremely useful, been having issues that relate to this.  Cheers!!

Fuseteam:
Still useful

README.md
FCLC: Missed this. 

You could, but there's no real cost to having it in primary dependencies. 

at worse apt/pacman etc. just let's the user know it's already at the latest version, same as any other deps

FCLC:
Request new review and potential up-streaming.

Issue #928, opened 2018-09-28
Full Screen issue, Exit button issue. 


leewp14:
1. Please check that no similar bug is already reported. Have a look on the list of open bugs at https://github.com/anbox/anbox/issues
2. Make sure you are running the latest version of Anbox before reporting an issue. Update snap to latest: `snap refresh --devmode --edge anbox`
3. Make sure you have debug logs enabled:
`sudo snap set anbox debug.enable=true`
4. Reproduce the error while debug logs enabled.
5. Run the anbox logs collection utility and attach the tar file.
`sudo /snap/bin/anbox.collect-bug-info `
[anbox-system-diagnostics-2018-09-28.zip](https://github.com/anbox/anbox/files/2427713/anbox-system-diagnostics-2018-09-28.zip)

6. ** Please paste the result of `anbox system-info` below:**
```
leewp14@CTK942S:~/Desktop$ anbox system-info
version: 4
snap-revision: 157
cpu:
  arch:  x86
  brand: Intel(R) Core(TM) i7 CPU       M 620  @ 2.67GHz
  features: 
    - aes
os:
  name: Ubuntu
  version: 16.04.2 LTS (Xenial Xerus)
  snap-based: true
kernel:
  version: Linux version 4.13.0-32-generic (buildd@lgw01-amd64-004) (gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.5)) #35~16.04.1-Ubuntu SMP Thu Jan 25 10:13:43 UTC 2018
  binder: true
  ashmem: true
graphics:
  egl:
    vendor: Mesa Project
    version: 1.4 (DRI2)
    extensions:
      - EGL_CHROMIUM_sync_control
      - EGL_EXT_image_dma_buf_import
      - EGL_KHR_cl_event2
      - EGL_KHR_config_attribs
      - EGL_KHR_create_context
      - EGL_KHR_create_context_no_error
      - EGL_KHR_fence_sync
      - EGL_KHR_get_all_proc_addresses
      - EGL_KHR_gl_colorspace
      - EGL_KHR_gl_renderbuffer_image
      - EGL_KHR_gl_texture_2D_image
      - EGL_KHR_gl_texture_3D_image
      - EGL_KHR_gl_texture_cubemap_image
      - EGL_KHR_image
      - EGL_KHR_image_base
      - EGL_KHR_image_pixmap
      - EGL_KHR_no_config_context
      - EGL_KHR_reusable_sync
[...]

PR #1810, opened 2021-05-13
fix the issue that the title bar disappears under multi-window strategy in some specific cases

### LINKED ISSUE #1184: Title bar disappearing, causing app usage issues
I'm having an issue with a spesific app, where the top title bar disappears when I click near the top of the screen. I then have to wiggle the window to get the top bar to render again. When this happens, a black bar appears at the bottom, the same size as the title bar. Additionally, all the buttons of the title bar still work when it isn't visisble. This isn't a new issue as it was first reported on issue #928. If log files are the likes are needed, they can be found there.

Here's what it should look like:
image
Here's what happens when affected:
image

swwind:
maybe fix #1184 #928

After some tests, I found that there are two situations cause the vanishment of the title bar.

One of the situation is that the `renderable` of the title bar only return the size of the title bar itself (which SHOULD be the whole size of the app). Original algorithm only find the max-sized rectangle and could not handle this situation correctly.

And another situation is a little weird. The `renderable` of the title bar just vanished without any clue, and the reason of this is beyond my knowledge.

This PR only fix the first situation. But the second situation could be easily turned to the first situation by moving the window around.

WARNING: only tested a few APKs and only tested on my laptop with kde, archlinux.

Good Luck

### DIFF
--- a/src/anbox/graphics/multi_window_composer_strategy.cpp
+++ b/src/anbox/graphics/multi_window_composer_strategy.cpp
@@ -47,19 +47,21 @@ std::map<std::shared_ptr<wm::Window>, RenderableList> MultiWindowComposerStrateg
   for (auto &w : win_layers) {
     const auto &renderables = w.second;
     RenderableList final_renderables;
-    auto new_window_frame = Rect::Invalid;
-    auto max_layer_area = -1;
+
+    int32_t new_left   = INT_MAX;
+    int32_t new_top    = INT_MAX;
+    int32_t new_right  = INT_MIN;
+    int32_t new_bottom = INT_MIN;
 
     for (auto &r : renderables) {
-      const auto layer_area = r.screen_position().width() * r.screen_position().height();
       // We always prioritize layers which are lower in the list we got
       // from SurfaceFlinger as they are already ordered.
-      if (layer_area < max_layer_area)
-        continue;
-
-      max_layer_area = layer_area;
-      new_window_frame = r.screen_position();
+      if (r.screen_position().left()   < new_left)   new_left   = r.screen_position().left();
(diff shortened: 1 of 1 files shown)


src/anbox/graphics/multi_window_composer_strategy.cpp
```diff
@@ -47,19 +47,21 @@ std::map<std::shared_ptr<wm::Window>, RenderableList> MultiWindowComposerStrateg
...
     for (auto &r : renderables) {
-      const auto layer_area = r.screen_position().width() * r.screen_position().height();
       // We always prioritize layers which are lower in the list we got
       // from SurfaceFlinger as they are already ordered.
-      if (layer_area < max_layer_area)
-        continue;
-
-      max_layer_area = layer_area;
-      new_window_frame = r.screen_position();
+      if (r.screen_position().left()   < new_left)   new_left   = r.screen_position().left();
```
morphis, line 59: Can you inherit the code style here the rest of the code uses? 

```
if (r.screen_position().left() < new_left) 
  new_left = r.screen_position().left();
```

morphis (changes requested):

src/anbox/graphics/multi_window_composer_strategy.cpp
swwind: ok.

morphis (approved):
Thanks! LGTM

Fuseteam:
@morphis i think you forgot to merge this one

soshimee:
@morphis hey?

Ruakker:
@morphis could you please merge this?

Fuseteam:
Changes still approved

swwind:
This PR is one year old now. :partying_face: :cake:

~~But nothing has changed.~~

PR #1837, opened 2021-06-19
Missing include for std::uint32_t

Benni3D:
Found while trying to build on Manjaro.


Fuseteam:
Doesn't sound like it should be closed
🤔

necessarily-equal (approved):

PR #1848, opened 2021-07-01
Add --launch-bounds parameter to anbox launch

firemeteorxx:
Also fix the following at server side
1. The API stub used an Invalid rect as launching bound.
2. The Android side service emitted the launch bound as one arg while
   'am start' expect four args.

### DIFF
--- a/android/service/android_api_skeleton.cpp
+++ b/android/service/android_api_skeleton.cpp
@@ -82,12 +82,10 @@ void AndroidApiSkeleton::launch_application(anbox::protobuf::bridge::LaunchAppli
 
     if (request->has_launch_bounds()) {
         argv.push_back("--launch-bounds");
-        std::stringstream launch_bounds;
-        launch_bounds << request->launch_bounds().left() << " "
-                      << request->launch_bounds().top() << " "
-                      << request->launch_bounds().right() << " "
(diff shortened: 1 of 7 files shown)


mcarans:
Would be great to see this merged

mhoffmann75:
For me no app window appears at all as soon as i add **--launch-bound=** to **anbox launch** command and no error message. Any perquisites that may be missing?
Omitting launch-bounds makes the apps open. Also tried several apps, and session-manager with CSD and SSD but does not change anything. 
Any idea?

firemeteorxx:
> For me no app window appears at all as soon as i add **--launch-bounds=** to **anbox launch** command and no error message. Any perquisites that may be missing? Omitting launch-bounds makes the apps open. Also tried several apps, and session-manager with CSD and SSD but does not change anything. Any idea?

It has been a while since I worked on this so I can't figure out what's the problem immediately.
Just a reminder that the change needs a corresponding Android OS image rebuild to work properly.
Are you suffering from the problem with or without OS image rebuild?

mhoffmann75:
Thanks for pointing that out. I must have missed the fact that i need a modified Android image. So i tried it with the original Image that came with anbox. So it's obvious that it cannot work :-)

Issue #1852, opened 2021-07-02
Anbox with apparmor settings?


stevenvandenbrandenstift:
Hello,

on a distro that does by default not use apparmor or snaps (alpinelinux),
the apparmor setting is breaking the startup.

For now we tried just making a patch to disable the setting for alpine and that makes the container start.
https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/22712/diffs#78433bd8a2455730339f7eb48121ce83cd93b235

Can I make a pull request with a extra ifdef cause to disable the setting (to disable apparmor support) for the container or do you have other suggestions?

necessarily-equal:
Hi, I went ahead and made a pull request for it https://github.com/anbox/anbox/pull/1856. By experience, I'd say not much will happen by making issues on this bug tracker :sweat_smile:

stevenvandenbrandenstift:
Well, I believe the pull request will never be accepted since its probably needed if you do use apparmor, I think we need a compile time option to disable apparmor support. Lets see if we can get response on here or pull request.

Fuseteam:
Has an open pull request

PR #1856, opened 2021-07-08
lxc: remove apparmor dependency

necessarily-equal:
asking for "unconfined" apparmor profile causes a crash with kernels
that don't have apparmor support enabled.

> Failed to start container: Failed to start container: Failed to set config item lxc.apparmor.profile)

Closes https://github.com/anbox/anbox/issues/1852.


necessarily-equal:
still useful

Issue #1735, opened 2021-01-22
Not able to run anbox on ubuntu 20.04


Aditya1927:
sudo add-apt-repository ppa:morphis/anbox-support
 
 More info: https://launchpad.net/~morphis/+archive/ubuntu/anbox-support
Press [ENTER] to continue or Ctrl-c to cancel adding it.

Hit:1 http://dl.google.com/linux/chrome/deb stable InRelease
Hit:2 http://archive.ubuntu.com/ubuntu focal InRelease                         
Hit:3 http://archive.ubuntu.com/ubuntu focal-updates InRelease                 
Ign:4 http://ppa.launchpad.net/morphis/anbox-support/ubuntu focal InRelease
Hit:5 http://archive.ubuntu.com/ubuntu focal-backports InRelease         
Hit:6 http://archive.ubuntu.com/ubuntu focal-security InRelease          
Err:7 http://ppa.launchpad.net/morphis/anbox-support/ubuntu focal Release
  404  Not Found [IP: 91.189.95.85 80]
Reading package lists... Done
E: The repository 'http://ppa.launchpad.net/morphis/anbox-support/ubuntu focal Release' does not have a Release file.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.

jnnksdev:
The apt package is outdated unfortunately.
You can use the snap package or build it from source.

calidion:
@jnnksdev snap is not running either.

nicolalandro:
I build [anbox-modules] and seams it works. I also build anbox. but it does not work the error is on the container:
```
$ sudo anbox container-manager                                       10:10:02
[sudo] password di xxx:          
[ 2021-01-29 09:10:13] [container_manager.cpp:74@operator()] You are running the container manager manually which is most likely not
[ 2021-01-29 09:10:13] [container_manager.cpp:75@operator()] what you want. The container manager is normally started by systemd or
[ 2021-01-29 09:10:13] [container_manager.cpp:76@operator()] another init system. If you still want to run the container-manager
[ 2021-01-29 09:10:13] [container_manager.cpp:77@operator()] you can get rid of this warning by starting with the --daemon option.
[ 2021-01-29 09:10:13] [container_manager.cpp:78@operator()] 
[ 2021-01-29 09:10:13] [container_manager.cpp:122@operator()] boost::filesystem::create_directories: Invalid argument
```

[...]

Issue #123, opened 2017-04-16
Failed to connect to socket /run/anbox-container.socket: No such file or directory


Connor124:
**System**:
Running kernel: 4.9.0-2-amd64
Distro: Debian 9
Anbox: 0.1.0 ( The most recent at time of writing. )

**Problem:**
Binder_Linux was compiled from source because the precompiled version didn't play nice with my system.
Anbox was installed via the anbox-installer script that was modified slightly to allow installation on unsupported systems. (I added "Debian" in the supported distros list)
When I run `anbox session-manager` I get `Failed to connect to socket /run/anbox-container.socket: No such file or directory`
I have tried "reverse-engineering" the installer script and manually compiling everything.

Any help is appreciated.

gasinvein:
Is container-manager running?

Connor124:
No, `Android image does not exist at path "/var/lib/anbox/android.img"`

gasinvein:
So try to place [image](http://build.anbox.io/android-images/2017/04/12/android_1_amd64.img) there, run `sudo anbox container-manager`, then `anbox session-manager`.

Connor124:
Does the image need special permissions, etc. Because, it is still not found.

gasinvein:
Maybe you should post there #114

Connor124:
I posted a solution to #114. But now, of course, there's another error that happens when I run anbox launch: `Stop waiting as we're already waiting for too long. Something is wrong with your setup and the container may have failed to boot.`

FHFS:
After fixing binder, i am running into the same problem.
> [fhfs@fhfs-Arch anbox-binder]$ anbox session-manager
[EE 2017-06-03 17:10:46] [daemon.cpp:58@Run] Failed to connect to socket /run/anbox-container.socket: No such file or directory

I do have the image.
> [fhfs@fhfs-Arch anbox-binder]$ ls /var/lib/anbox/
android.img

But i am unable to run container manager.
> [fhfs@fhfs-Arch anbox-binder]$ sudo anbox container-manager
[WW 2017-06-03 17:12:12] [container_manager.cpp:58@operator()] You are running the container manager manually which is most likely not
[WW 2017-06-03 17:12:12] [container_manager.cpp:59@operator()] what you want. The container manager is normally started by systemd or
[WW 2017-06-03 17:12:12] [container_manager.cpp:60@operator()] another init system. If you still want to run the container-manager
[...]

PR #1870, opened 2021-07-19
container-manager: fix create_directories error

spholz:
anbox container-manager crashes if `data_path_` of `anbox::cmds::ContainerManager` is not set with the error message `boost::filesystem::create_directories: Invalid argument` (#123, #1735 seem to have the same issue)
It looks like container-manager crashes on some systems if an empty path is passed to `fs::create_directories` in `src/anbox/cmds/container_manager.cpp:99`.

This PR adds an additional check to make sure the path is not empty.

### DIFF
--- a/src/anbox/cmds/container_manager.cpp
+++ b/src/anbox/cmds/container_manager.cpp
@@ -95,7 +95,7 @@ anbox::cmds::ContainerManager::ContainerManager()
       if (!data_path_.empty())
         SystemConfiguration::instance().set_data_path(data_path_);
 
-      if (!fs::exists(data_path_))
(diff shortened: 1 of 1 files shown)


morphis (approved):
LGTM

Fuseteam:
Waiting for the build

spholz:
The "checks" tab is empty but the state of the build task is "Expected — Waiting for status to be reported"
I think when I originally created this PR the build task didn't start because this repo limited automatic builds to people who already commited to this repo. But now the job is missing?
How can I fix this?

Fuseteam:
I think @morphis should be able to trigger it too :thinking:

PR #1875, opened 2021-07-28
emugl:fix webview,GL_SHADING_LANGUAGE_VERSION and EGL_RENDERABLE_TYPE

kuailexs:
Webview will try to use Opengles3, but Anbox does not support it.

GetGLString(GL_SHADING_LANGUAGE_VERSION)，
eglChooseConfig()

see:
https://github.com/chromium/chromium/blob/67.0.3396.87/gpu/config/gpu_info_collector.cc#L160
https://github.com/chromium/chromium/blob/67.0.3396.87/ui/gl/gl_surface_egl.cc#L330


kuailexs:
will make webview running,test prebuilt  67.0.3396.87 and 52.0.2743.100

图片

PR #1887, opened 2021-08-18
Anbox WebCam 

mister5111:
I use this manual  https://github.com/alecbarber/anbox 

Does someone have a camera?


ahmubashshir (commented):
You shouldn't create PRs from master, do something like this
```sh
git checkout upstream/master
git checkout -b webcam
git cherry-pick <relevant-commit-sha>
```
where those commit sha would be commits in this pr except this one commit.
Then update Readme.md w/o removing the upstream stuffs.

PR #1924, opened 2021-10-05
snap: fix failure on cloning SwiftShader repository

memnoth:
snapcraft uses '--recursive-submodules' option by default to clone all
dependencies if source-type is 'git'. Therefore the option leads the failure
on cloning the project by 'git-hooks' submodule. In the case, the failure is
asking username/password to do so.

Use 'override-pull' to prevent the issue as the following
1. clone the project w/o submodules.
2. use 'sed' command to remove 'git-hooks' submodule.
3. update submodule

Fix: https://github.com/anbox/anbox/issues/1922


snap/snapcraft.yaml
```diff
@@ -209,9 +209,15 @@ parts:
 
   swiftshader:
     plugin: cmake
-    source: https://swiftshader.googlesource.com/SwiftShader
-    source-type: git
-    source-commit: b6e8c3f0f4830887d69ba765a922ac3c40e81dd9
+    build-packages:
```
memnoth, line 212: I am not sure this is requirement even though 'sed' is used for only 'pull'.

PR #1938, opened 2021-11-04
Work around bug in lxc 4.0.10

Skirmisher:
lxc 4.0.10 evidently has [an issue](https://github.com/lxc/lxc/issues/3964) where specifying `sys.mixed` in the value of `lxc.mount.auto` causes it to fail to bind-mount `/sys/devices/virtual/net` for some reason, only leaving the unhelpful message `Operation not permitted - Failed to mount "(null)" onto "/usr/lib64/lxc/rootfs/sys/devices/virtual/net"` in `/var/lib/anbox/logs/container.log`. Specifying `sys.ro` instead of `sys.mixed` and then adding an explicit bind mount for the problematic path (i.e. replicating the behavior of `sys.mixed`) works around the problem.

### DIFF
--- a/src/anbox/container/lxc_container.cpp
+++ b/src/anbox/container/lxc_container.cpp
@@ -336,7 +336,9 @@ void LxcContainer::start(const Configuration &configuration) {
 
   // We can mount proc/sys as rw here as we will run the container unprivileged
   // in the end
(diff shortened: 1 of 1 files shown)


Fuseteam:
Still sounds useful

PR #1827, opened 2021-06-07
Fix typo in lxc.cgroup config breaking lxc >= 4.0.9

pablofsf:
Closes #1801

LXC 4.0.9 is a LTS and should not have any breaking change in behavior. I believe that the issue in #1801 is due to a typo in the original configuration. At least since version 2.0.0 (I checked the docs in lxc for the corresponding tag),  `lxc.group.devices` has not been a valid configuration, but `lxc.cgroup.devices` instead. I guess there was some bug that silently ignored that invalid configuration previously to  4.0.9

PR #1998, opened 2022-02-12
Fix compilation on openSUSE

leviathanch:
Hi
On openSUSE those small changes are needed for successfully building an RPM.
Please consider pulling this patch.
Thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,8 @@ pkg_check_modules(LXC lxc REQUIRED)
 pkg_check_modules(PROPERTIES_CPP properties-cpp REQUIRED)
 pkg_check_modules(LIBSYSTEMD libsystemd REQUIRED)
 pkg_check_modules(LIBCAP libcap REQUIRED)
+pkg_check_modules(GTEST_BOTH gtest_main REQUIRED)
+pkg_check_modules(GMOCK gmock_main REQUIRED)
 
 message(STATUS "LXC version: ${LXC_VERSION}")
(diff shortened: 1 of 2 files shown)


Fuseteam:
@leviathanch so those camera commits aren't intented for this PR?

leviathanch:
Nope. Just really only that one patch I've proposed to merge, so that RPMs 
build through on openSUSE...
The camera support still is buggy and it turns out that's because of some
more serious issue with the Hudini libraries which cause an "has unexpected 
e_machine: 183" error right now, because cross platform support with binfmt
isn't working yet and that messes up the entire process of accessing the 
camera for some reason... and Revolut doesn't start, which was the only
reason I went onto that endeavour.

Cheers
-lev

Fuseteam:
the camera commits ended up as a part of this PR, you may want to remove them

leviathanch:
Oops...

leviathanch:
Fixed it

Fuseteam:
Is this commit also intended to be part of this PR https://github.com/anbox/anbox/pull/1998/commits/000df0d8bb44481f56907ca1003687f4ae0743ec @leviathanch?

leviathanch:
Yeah. Because otherwise Anbox doesn't work with the current LXC version openSUSE

Fuseteam:
That sounds more like this PR depends on #1827.
hmmm at this point it's up to @morphis how he wants to handle it

morphis (approved):
LGTM

PR #2029, opened 2022-04-19
Use std::{make_shared,make_unique}

necessarily-equal:


src/anbox/common/loop_device.cpp
```diff
@@ -31,7 +31,7 @@ std::shared_ptr<LoopDevice> LoopDevice::create(const boost::filesystem::path &pa
   if (fd < 0)
     throw std::system_error{errno, std::system_category()};
 
-  return std::shared_ptr<LoopDevice>(new LoopDevice(Fd{fd}, path));
+  return std::make_shared<LoopDevice>(Fd{fd}, path);
```
morphis, line 34: The c'tor isn't public so this wont be possible. You have to leave this as `std::shared_ptr<LoopDevice>(new LoopDevice(Fd{fd}, path));`

morphis (changes requested):
Most of the changes wont work given that the c'tors are private for a good reason. What's the purpose of this change?

PR #2030, opened 2022-04-19
Fix binderfs detection

necessarily-equal:


src/anbox/cmds/system_info.cpp
```diff
@@ -162,7 +161,7 @@ class SystemInformation {
     }
 
     kernel_info_.binder = fs::exists(binder_path);
-    kernel_info_.binderfs = fs::exists(binderfs_path);
+    kernel_info_.binderfs = common::BinderDeviceAllocator::is_supported();
```
morphis, line 164: @necessarily-equal the `common` namespace doesn't seem to be available and that makes CI fail. Can you fix that?

morphis (changes requested):

PR #2085, opened 2022-09-07
Compatibility for anbox-bridge.sh with systemd.generator

Matteljay:
The comments instruct systemd on the correct boot order for anbox-bridge.sh


morphis (approved):
LGTM, thanks!

